PR #356, opened 2026-03-19
Axi4 passive vip and scoreboard

csabakiss-semify:
1. Developed AXI4 VIP supporting passive mode only
2. VIP integrated to the top level environment
3. Added scoreboard to the AXI XBAR


hw/top_chip/dv/tb/tb.sv
martin-velay: Here we should follow this rule: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#preferred-systemverilog-constructs


```suggestion
  always_comb begin
```

PR #366, opened 2026-03-23
util: FPGA runner: rework to use asyncio

ziuziakowska:
Split off from: https://github.com/lowRISC/mocha/pull/299

This reworks the FPGA runner to use asyncio to read from the UART while the flash is being written. This was causing some problems for me locally when adding some more debugging output, as in the time between finishing flashing and opening the UART file descriptor, the receive buffer would be overflowed and the test status would be dropped.

### DIFF
--- a/util/fpga_runner.py
+++ b/util/fpga_runner.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
-import subprocess
+import asyncio
 import sys
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


util/fpga_runner.py
```diff
@@ -17,55 +17,43 @@
...
-                continue
+async def load_fpga_test(test: Path) -> None:
+    command = ["openFPGALoader", "--spi", "--offset", str(BOOT_ROM_OFFSET), "--write-flash"]
+    command.append(str(test.with_suffix(".bin")))
 
-            if "PASSED" in line:
-                return 0
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
+        sys.exit(1)
```
engdoreis: If the load fails we would not know why? I suggest keeping the log.

PR #382, opened 2026-03-26
[chip, gpio, dv] GPIO smoke test

### LINKED ISSUE #147: GPIO DV - Top-level smoke test
- [x] Basic reg access in C for Verilator
- [x] Smoke test (reg+port activity) for top UVM test

KinzaQamar:
This PR added chip level GPIO smoke test.

Will close: https://github.com/lowRISC/mocha/issues/147

### DIFF
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_smoke_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_smoke_vseq.sv
@@ -1,0 +1,52 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class top_chip_dv_gpio_smoke_vseq extends top_chip_dv_base_vseq;
+  `uvm_object_utils(top_chip_dv_gpio_smoke_vseq)
+
+  // Standard SV/UVM methods
+  extern function new(string name="");
+  extern task body();
+
+  // Class specific methods
+  //
+  // Wait for the pattern to appear on the GPIO's pads
+  extern virtual task wait_for_pattern(logic [NUM_GPIOS-1:0] exp_val);
+endclass : top_chip_dv_gpio_smoke_vseq
+
+function top_chip_dv_gpio_smoke_vseq::new (string name = "");
+  super.new(name);
+endfunction : new
+
+task top_chip_dv_gpio_smoke_vseq::body();
+  super.body();
+  `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
+
+  `uvm_info(`gfn, "Starting GPIOs outputs test", UVM_LOW)
+
+  // Disable pin pull-ups enabled to support SW booting process
+  cfg.gpio_vif.set_pullup_en('0);
+
+  // SW first drives walking 1's on each pin. Wait till those patterns are visible
+  for (int i = 0; i < NUM_GPIOS; i++) begin
+    wait_for_pattern(1 << i);
+  end
+
+  // Wait for Z so that the pads can safely be driven in inputs
+  wait_for_pattern('Z);
+
+  // The outputs on the pads are seen on the edge of sys_clk_if.clk so wait at least a negedge
+  // before driving the pads in inputs.
+  cfg.sys_clk_vif.wait_n_clks(1);
+
+  `uvm_info(`gfn, "Starting GPIOs inputs test", UVM_LOW)
+
+  // Drive the pads as inputs
+  cfg.gpio_vif.drive('h5555_5555);
+endtask : body
+
+task top_chip_dv_gpio_smoke_vseq::wait_for_pattern(logic [NUM_GPIOS-1:0] exp_val);
+  `DV_SPINWAIT(wait(cfg.gpio_vif.pins === exp_val);,
+               $sformatf("Timed out waiting for GPIOs == %0h", exp_val))
+endtask : wait_for_pattern
--- a/hw/top_chip/dv/env/top_chip_dv_env.core
+++ b/hw/top_chip/dv/env/top_chip_dv_env.core
@@ -23,6 +23,7 @@ filesets:
       - seq_lib/top_chip_dv_vseq_list.sv: {is_include_file: true}
       - seq_lib/top_chip_dv_base_vseq.sv: {is_include_file: true}
       - seq_lib/top_chip_dv_uart_base_vseq.sv: {is_include_file: true}
+      - seq_lib/top_chip_dv_gpio_smoke_vseq.sv: {is_include_file: true}
     file_type: systemVerilogSource
 
 targets:
--- a/hw/top_chip/dv/env/top_chip_dv_env.core
+++ b/hw/top_chip/dv/env/top_chip_dv_env.core
@@ -12,6 +12,7 @@ filesets:
       - lowrisc:dv:mem_bkdr_util
       - lowrisc:dv:uart_agent
       - lowrisc:dv:common_ifs
+      - lowrisc:mocha_dv:gpio_env
     files:
       - top_chip_dv_env_pkg.sv
       - mem_clear_util.sv: {is_include_file: true}
--- a/hw/top_chip/dv/tb/tb.sv
+++ b/hw/top_chip/dv/tb/tb.sv
@@ -10,6 +10,7 @@ module tb;
   import mem_bkdr_util_pkg::mem_bkdr_util;
   import top_chip_dv_env_pkg::*;
   import top_chip_dv_test_pkg::*;
+  import gpio_env_pkg::NUM_GPIOS;
 
   import top_chip_dv_env_pkg::SW_DV_START_ADDR;
   import top_chip_dv_env_pkg::SW_DV_TEST_STATUS_ADDR;
@@ -26,6 +27,11 @@ module tb;
   wire peri_clk;
   wire peri_rst_n;
 
+  // GPIO connections
+  wire  [NUM_GPIOS-1:0] gpio_pads;    // A wire connected to bidirectional pads in pins_if
+  logic [NUM_GPIOS-1:0] dut_gpio_o;
+  logic [NUM_GPIOS-1:0] dut_gpio_en_o;
+
   logic [3:0] spi_host_sd;
   logic [3:0] spi_host_sd_en;
 
@@ -33,6 +39,7 @@ module tb;
   clk_rst_if sys_clk_if(.clk(clk), .rst_n(rst_n));
   clk_rst_if peri_clk_if(.clk(peri_clk), .rst_n(peri_rst_n));
   uart_if uart_if();
+  pins_if #(NUM_GPIOS) gpio_pins_if (.pins(gpio_pads));
 
   // ------ Mock DRAM ------
   top_pkg::axi_dram_req_t  dram_req;
@@ -55,6 +62,10 @@ module tb;
     // Clock and reset.
     .clk_i                (clk              ),
     .rst_ni               (rst_n            ),
+    // GPIO inputs and outputs with output enable
+    .gpio_i               (gpio_pads        ),
+    .gpio_o               (dut_gpio_o       ),
+    .gpio_en_o            (dut_gpio_en_o    ),
     // UART receive and transmit.
     .uart_rx_i            (uart_if.uart_rx  ),
     .uart_tx_o            (uart_if.uart_tx  ),
@@ -88,6 +99,12 @@ module tb;
     .dram_req_o           (dram_req         ),
     .dram_resp_i          (dram_resp        )
   );
+
+  // Assignment to the GPIO pads. If dut_gpio_en_o[i] is disabled, then let the gpio_pad[i] float so
+  // an external device / driver can drive it.
+  for (genvar i = 0; i < NUM_GPIOS; i++) begin : gen_gpio_pads
+    assign gpio_pads[i] = dut_gpio_en_o[i] ? dut_gpio_o[i] : 1'bz;
+  end
 
   // Signals to connect the sink
   top_pkg::axi_req_t  sim_sram_cpu_req;
@@ -204,6 +221,10 @@ module tb;
     `SIM_SRAM_IF.sw_dv_size                               = SW_DV_SIZE;
     `SIM_SRAM_IF.u_sw_test_status_if.sw_test_status_addr  = SW_DV_TEST_STATUS_ADDR;
     `SIM_SRAM_IF.u_sw_logger_if.sw_log_addr               = SW_DV_LOG_ADDR;
+
+    // Enable GPIO pull-ups to support the SW boot process. While only pin 8 is strictly required
+    // by the Boot ROM to determine the boot path, we enable pull-ups on all pins for consistency.
+    gpio_pins_if.set_pullup_en('1);
 
     // Start clock and reset generators
     sys_clk_if.set_active();
--- a/hw/top_chip/dv/tb/tb.sv
+++ b/hw/top_chip/dv/tb/tb.sv
@@ -233,6 +233,7 @@ module tb;
     uvm_config_db#(virtual clk_rst_if)::set(null, "*", "sys_clk_if", sys_clk_if);
     uvm_config_db#(virtual clk_rst_if)::set(null, "*", "peri_clk_if", peri_clk_if);
     uvm_config_db#(virtual uart_if)::set(null, "*.env.m_uart_agent*", "vif", uart_if);
+    uvm_config_db#(virtual pins_if #(NUM_GPIOS))::set(null, "*.env", "gpio_vif", gpio_pins_if);
 
     // SW logger and test status interfaces.
     uvm_config_db#(virtual sw_test_status_if)::set(
--- a/hw/top_chip/dv/top_chip_sim_cfg.hjson
+++ b/hw/top_chip/dv/top_chip_sim_cfg.hjson
@@ -142,7 +142,7 @@
     }
     {
       name: gpio_smoke
-      uvm_test_seq: top_chip_dv_gpio_vseq
+      uvm_test_seq: top_chip_dv_gpio_smoke_vseq
       sw_images: ["gpio_smoketest_vanilla_bare:5"]
       run_opts: ["+ChipMemSRAM_image_file={run_dir}/gpio_smoketest_vanilla_bare.vmem"]
     }
--- a/hw/top_chip/dv/top_chip_sim_cfg.hjson
+++ b/hw/top_chip/dv/top_chip_sim_cfg.hjson
@@ -147,6 +147,12 @@
       run_opts: ["+ChipMemSRAM_image_file={run_dir}/gpio_smoketest_vanilla_bare.vmem"]
     }
     {
+      name: gpio_smoke_cheri
+      uvm_test_seq: top_chip_dv_gpio_smoke_vseq
+      sw_images: ["gpio_smoketest_cheri_bare:5"]
+      run_opts: ["+ChipMemSRAM_image_file={run_dir}/gpio_smoketest_cheri_bare.vmem"]
+    }
+    {
       name: rom_ctrl_integrity_check
       uvm_test_seq: top_chip_dv_rom_ctrl_integrity_check_vseq
       sw_images: ["rom_ctrl_integrity_check_test_vanilla_bare:5"]
@@ -249,6 +255,7 @@
         "spi_host_smoke",
         "spi_host_smoke_cheri",
         "gpio_smoke",
+        "gpio_smoke_cheri",
         "mailbox_smoke",
         "mailbox_smoke_cheri",
         "axi_sram_smoke",
@@ -304,7 +311,8 @@
     {
       name: gpio
       tests: [
-        "gpio_smoke"
+        "gpio_smoke",
+        "gpio_smoke_cheri"
       ]
     }
     {
--- a/sw/device/tests/CMakeLists.txt
+++ b/sw/device/tests/CMakeLists.txt
@@ -9,6 +9,9 @@ mocha_add_test(NAME test_framework_exception_test SOURCES test_framework/excepti
 mocha_add_test(NAME axi_sram_smoketest SOURCES axi_sram/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME axi_sram_tag_test SOURCES axi_sram/tag_test.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME gpio_reg_access_test SOURCES gpio/reg_access_test.c LIBRARIES ${LIBS} FPGA)
+# Can't run this test on FPGA or verilator tops and this test thinks that the GPIO inputs can be
+# driven externally
+mocha_add_test(NAME gpio_smoketest SOURCES gpio/smoketest.c LIBRARIES ${LIBS} SKIP_VERILATOR)
 mocha_add_test(NAME i2c_smoketest SOURCES i2c/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME mailbox_smoketest SOURCES mailbox/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME plic_smoketest SOURCES plic/smoketest.c LIBRARIES ${LIBS} FPGA)
--- a/sw/device/tests/gpio/smoketest.c
+++ b/sw/device/tests/gpio/smoketest.c
@@ -1,0 +1,45 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "hal/gpio.h"
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+static bool gpio_test(gpio_t gpio)
+{
+    uint8_t timeout = 0;
+
+    // Enable the GPIOs in output mode
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0xFFFFFFFF);
+
+    // Set each pin in walking 1's fashion
+    for (int i = 0; i < GPIO_NUM_PINS; i++) {
+        DEV_WRITE(gpio + GPIO_REG_DIRECT_OUT, 1 << i);
+    }
+
+    // Once the SW finishes driving the walking 1's pattern on the last GPIO pin, it should disable
+    // the output enables so that the GPIO pads can safely be driven as inputs
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0x0);
+
+    // Wait for the pattern of 1's on odd pins and 0's on even pins to be driven externally.
+    while (DEV_READ(gpio + GPIO_REG_DATA_IN) != 0x55555555) {
+        // The expected pattern is driven externally on the inputs by toggling the wires. Since this
+        // occurs without delay in simulation, a timeout of 10 read requests is sufficient to detect
+        // a failure.
+        if (timeout == 0xA) {
+            return false;
+        }
+        timeout++;
+    }
+
+    return true;
+}
+
+bool test_main()
+{
+    gpio_t gpio = mocha_system_gpio();
+    return gpio_test(gpio);
+}
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_vseq_list.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_vseq_list.sv
@@ -4,3 +4,4 @@
 
 `include "top_chip_dv_base_vseq.sv"
 `include "top_chip_dv_uart_base_vseq.sv"
+`include "top_chip_dv_gpio_smoke_vseq.sv"
--- a/hw/top_chip/dv/env/top_chip_dv_env.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_env.sv
@@ -44,6 +44,11 @@ function void top_chip_dv_env::build_phase(uvm_phase phase);
     `uvm_fatal(`gfn, "failed to get sw_logger_vif from uvm_config_db")
   end
 
+  // Get the GPIO VIF handle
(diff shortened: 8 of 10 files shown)


martin-velay (commented):
I have some initial comments. It's a very good start though!

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv
```diff
@@ -0,0 +1,130 @@
...
+  super.new(name);
+endfunction : new
+
+task top_chip_dv_gpio_base_vseq::body();
+  super.body();
+  `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
+  gpio_test();
+endtask : body
+
+task top_chip_dv_gpio_base_vseq::drive_pattern(int unsigned num_clks, int unsigned pin, bit val);
```
martin-velay: Suggested name change:
```suggestion
task top_chip_dv_gpio_base_vseq::drive_pattern(int unsigned wait_clks, int unsigned pin, bit val);
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv
```diff
@@ -0,0 +1,130 @@
...
+  cfg.peri_clk_vif.wait_clks(num_clks);
+  cfg.gpio_vif.drive_pin(pin, val);
+endtask : drive_pattern
+
+task top_chip_dv_gpio_base_vseq::wait_for_pattern(logic [NUM_GPIOS-1:0] exp_val);
+  `DV_SPINWAIT(wait(cfg.gpio_vif.pins ==? exp_val);,
```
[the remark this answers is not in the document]
KinzaQamar: For this kind of comparison
```
  for (int i = 0; i < NUM_GPIOS/4; i++) begin
    wait_for_pattern({8'h?, 1 << i, 16'h????});
  end
  ```
  I want to compare the third quater of pins driven by software and ignore the rest.

File: hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+  // Check for walking 1's pattern on lower 8 pins.
+  for (int i = 0; i < NUM_GPIOS/4; i++) begin
+    wait_for_pattern({24'h?, 1 << i});
+  end
+
+  // GPIO pads : 'h00000080
+  //
+  // Drive 1's in temperature pattern on the lower 8 pins.
+  for (int i = NUM_GPIOS/4; i < (NUM_GPIOS/4) * 2; i++) begin
+    drive_pattern(1, i, 1);
```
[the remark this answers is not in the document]
KinzaQamar: Does that look fine now?

martin-velay (commented):
I feel that the choice of driving patterns on the quarters from the SW and DV at the same time and by crossing the quarters is hard to follow and I am not sure it brings something for a top-level test. I'd suggest to simplify but still driving and checking from both the SW and the DV sides. If you have a doubt we can have a call.

But we are almost there! 👌

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv
```diff
@@ -0,0 +1,131 @@
...
+  //
+  // Enable the pulldowns so that the pads are driving 0's rather than Z's when no external driver
+  // is connected
+  cfg.gpio_vif.set_pulldown_en('1);
+
+  // Current Current GPIOs pads state : 'h0
+
+  `uvm_info(`gfn, "Starting GPIOs outputs test", UVM_LOW)
+
+  // The SW first sets the first and third quater of GPIOs to 0's in order to walk 1's on them.
```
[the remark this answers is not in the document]
KinzaQamar: Maybe the best answer is to remove the `setting to 0's` thing from both SW and vseq. I was thinking it is good to drive 0's and wait for 0's then because the function can be used somewhere else in a different context and to do W1's cleanly, we should initialize it with 0's.

hw/top_chip/dv/env/top_chip_dv_env.core
```diff
@@ -12,6 +12,7 @@ filesets:
       - lowrisc:dv:mem_bkdr_util
       - lowrisc:dv:uart_agent
       - lowrisc:dv:common_ifs
+      - lowrisc:mocha_dv:gpio_env
```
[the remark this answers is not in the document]
KinzaQamar: No then the environment is unable to find `NUM_GPIOS`. There must be a dependency mess-up because I just saw in `build.log` that it is unable to find `top_chip_dv_env_pkg`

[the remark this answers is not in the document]
KinzaQamar: Well, I spent a while to make it work somehow. The problem atm is that I want to use `NUM_GPIOS` parameter from `gpio_env_pkg` in `tb.sv` and `uvm`. I have an import  `import gpio_env_pkg::NUM_GPIOS;` in `top_chip_dv_env_pkg` to be able to use that parameter in `UVM`. To achieve that I need to have `- lowrisc:mocha_dv:gpio_env` in `top_chip_dv_env.core`

martin-velay (approved):
Thanks Kinza, this PR is ready from my perspective!

> [!NOTE]
>  There is only the C test compatibility with Verilator to resolve as you've raised to me by chat (thanks for noticing). As it breaks the CI and as a feature is missing from the SW side, it's probably best to have another test for Verilator and open an issue to remember to use the same test later.

sw/device/lib/hal/gpio.h
```diff
@@ -25,5 +25,11 @@ bool gpio_read_pin(gpio_t gpio, uint32_t pin);
...
 void gpio_write_pin(gpio_t gpio, uint32_t pin, bool state);
 
+// Set the value of all output pins
+void gpio_write(gpio_t gpio, uint32_t output_reg_addr, uint32_t value);
+
 // Set the output enable state of one pin
 void gpio_set_oe_pin(gpio_t gpio, uint32_t pin, bool output);
+
+// Set the output enable state of all pins
+void gpio_set_all_oe(gpio_t gpio, uint32_t value);
```
elliotb-lowrisc: I think the consistency could be improved here in two ways:

1. Use the same ordering of the functions in "gpio.c" as "gpio.h". I don't mind which ordering you choose, just that they match.
2. Change the name of one of the new functions so they match each other. The original two functions both ended in `_pin`, but the new ones differ in how they convey that the affect multiple pins. Perhaps something like `gpio_set_all_oe` -> `gpio_set_oe` or `gpio_write` -> `gpio_write_all` (perhaps also with `gpio_set_all_oe` -> `gpio_set_oe_all`).

sw/device/lib/hal/gpio.c
```diff
@@ -30,3 +30,13 @@ void gpio_set_oe_pin(gpio_t gpio, uint32_t pin, bool output)
     uint32_t mask = 1u << (pin & 0xFu);
     DEV_WRITE(gpio + reg, (mask << 16) | (output ? mask : 0u));
 }
+
+void gpio_write(gpio_t gpio, uint32_t output_reg_addr, uint32_t value)
+{
+    DEV_WRITE(gpio + output_reg_addr, value);
+}
```
elliotb-lowrisc: At the moment I don't see a good reason to use this function instead of using `DEV_WRITE` directly. I'd suggest either specifying `GPIO_REG_DIRECT_OE` inside the function rather than requiring the caller to provide it (like `gpio_set_all_oe`), or removing this function entirely and just using `DEV_WRITE` in the smoketest (like how `DEV_READ` is used for gpio reads).

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv
```diff
@@ -17,5 +22,30 @@ endfunction : new
...
+  wait_for_pattern('Z);
+
+  // The outputs on the pads are seen on the edge of sys_clk_if.clk so wait at least a negedge
+  // before driving the pads in inputs.
+  cfg.sys_clk_vif.wait_n_clks(1);
+
+  `uvm_info(`gfn, "Starting GPIOs inputs test", UVM_LOW)
+
+  // Drive the pads as inputs
+  cfg.gpio_vif.drive({NUM_GPIOS{1'b1}});
```
elliotb-lowrisc: An even better pattern would be to include some zeros as well to make sure the inputs are not all stuck high

elliotb-lowrisc (changes requested):
Thanks for all your hard work on this and apologies I haven't reviewed this sooner. Here's my current feedback:

elliotb-lowrisc:
Ah, I see the verilator and FPGA tests are broken, perhaps the previous smoketest should remain and the new test you've created could be something else that is not run as part of the smoketests (seeing as it doesn't pass for verilator or FPGA).

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_base_vseq.sv
KinzaQamar: Good point. Now I've made odd pin high and even pin low

sw/device/lib/hal/gpio.c
KinzaQamar: I think you are right. I've drop that commit entirely.

KinzaQamar:
Thanks for your reveiw @elliotb-lowrisc. Yes, the smoketest was hanged forever as we don't have the SW to drive GPIO input's DPI port when using verilator top. I've made [406](https://github.com/lowRISC/mocha/pull/406) which is a dependency on this PR. That rename's your test with a `_verilator` suffix and then build will create *_verilator_* image. I am waiting for that to work on this PR in CI.

sw/device/tests/CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ set(LIBS hal runtime startup test_framework)
 
 mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME test_framework_exception_test SOURCES test_framework/exception.c LIBRARIES ${LIBS})
-mocha_add_test(NAME gpio_smoketest SOURCES gpio/smoketest.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME gpio_smoketest_verilator SOURCES gpio/smoketest_verilator.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME gpio_smoketest_xcelium SOURCES gpio/smoketest_xcelium.c LIBRARIES ${LIBS} FPGA)
```
elliotb-lowrisc: This is a list of tests to run in Verilator and optionally FPGA. I don't think an Xcelium-only test will work here

KinzaQamar: Yeah, you are right. I need to somehow tell the cmake to stop making verilator/FPGA images when the test contains `xcelium` in name.

KinzaQamar: Also, I think I can throw the FPGA option for Xcelium. I'll do it in the morning

sw/cmake/tests.cmake
engdoreis:
```suggestion
    set(options FPGA SIM_VERILATOR)
```

sw/cmake/tests.cmake
```diff
@@ -83,12 +83,16 @@ macro(mocha_add_test)
         # create artefacts
         mocha_add_executable_artefacts(${NAME})
 
-        if(SIM)
```
engdoreis:
```suggestion
        if(SIM AND arg_SIM_VERILATOR)
```

sw/device/tests/CMakeLists.txt
```diff
@@ -6,14 +6,14 @@ set(LIBS hal runtime startup test_framework)
 
 mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA)
```
engdoreis, line 7: The same for the other tests below. Except the one that can't run in verilator.
```suggestion
mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA SIM_VERILATOR)
```

hw/top_chip/dv/env/top_chip_dv_env.core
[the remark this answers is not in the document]
rswarbrick: Hmm. I *think* that the dependency chain should be:

1. The sim core contains `tb.sv`.
2. That sim core depends on the chip-level test.
3. The test depends on the chip-level environment.
4. The chip-level environment depends on the block-level environments for the various blocks.

But I just looked and I don't think `lowrisc:mocha_dv:top_chip_sim` depends on `lowrisc:mocha_dv:top_chip_dv_test`. This is a bug! (I'd probably suggest opening a 1-line PR to put in the dependency, which can land before this PR)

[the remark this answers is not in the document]
martin-velay: It might be a bug indeed, I struggled a bit to set the core files

sw/device/tests/CMakeLists.txt
```diff
@@ -4,20 +4,31 @@
...
-mocha_add_test(NAME plic_smoketest SOURCES plic/smoketest.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA
+               VERILATOR)
+mocha_add_test(NAME test_framework_exception_test SOURCES test_framework/exception.c LIBRARIES
+               ${LIBS} VERILATOR)
+mocha_add_test(NAME gpio_reg_access_test SOURCES gpio/reg_access_test.c LIBRARIES ${LIBS} FPGA
+               VERILATOR)
+# Can't run this test on FPGA and verilators top and this test thinks that the GPIO inputs can be
+# driven
+mocha_add_test(NAME gpio_smoke_test SOURCES gpio/smoke_test.c LIBRARIES ${LIBS} FPGA)
```
martin-velay: I just noticed the name of this test is different from all others, can you change it please?
`smoke_test` -> `smoketest`

KinzaQamar: I prefer `smoke_test` as both are different words but let's stick with the convention used in the file

engdoreis (commented):
The Software changes LGTM.

sw/device/tests/gpio/smoketest.c
```diff
@@ -0,0 +1,36 @@
...
+
+static bool gpio_test(gpio_t gpio)
+{
+    // Enable the GPIOs in output mode
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0xFFFFFFFF);
+
+    // Set each pin in walking 1's fashion
+    for (int i = 0; i < GPIO_NUM_PINS; i++) {
+        DEV_WRITE(gpio + GPIO_REG_DIRECT_OUT, 1 << i);
+    }
```
engdoreis, line 21: If you don't need multiple bus transactions.
```suggestion
    DEV_WRITE(gpio + GPIO_REG_DIRECT_OUT, (1 << GPIO_NUM_PINS) -1);
```

hw/top_chip/dv/env/top_chip_dv_env.core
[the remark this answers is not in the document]
KinzaQamar: Hi @rswarbrick,

I've created a PR for the GPIO dependency issue we discussed yesterday: [29769](https://github.com/lowRISC/opentitan/pull/29769)

This needs to go in first so that I'll vendor the stuff in mocha and then tweak this PR

sw/device/tests/gpio/smoketest.c
KinzaQamar: The idea is to see if we can set every output high through W1's. Pattern is going to be `1, 10, 100, 1000, ...`

marnovandermaas:
Converting to draft as it depends on #442

hw/top_chip/dv/tb/tb.sv
```diff
@@ -86,6 +95,12 @@ module tb;
     .dram_resp_i          (dram_resp        )
   );
 
+  // Assignment to the GPIO pads. If dut_gpio_en_o[i] is disabled, then let the gpio_pad[i] float so
+  // an external device/driver can drive it.
+  for (genvar i = 0; i < NUM_GPIOS; i++) begin : gen_gpio_pads
+    assign gpio_pads[i] = dut_gpio_en_o[i] ? dut_gpio_o[i] : 1'bz;
```
engdoreis, line 106: I just bumped into this when trying to make the bootrom boostrap pin to work with dvsim. Is it possible to make the pin 8 high if output enable high rathen then floating?

martin-velay: I guess it is possible, can you explain a bit more the reason in order to document this behavior with an explicit comment

engdoreis: The [gpio 8](https://github.com/lowRISC/mocha/blob/fac6f9b7046d92643ba1a157559f9371ac2ec667/hw/top_chip/data/pins_genesys2.xdc#L25) is now used by the bootrom to decide if it should enter spi bootstrap or jump to the next stage (test). For now we only want to enter spi bootstrap in the  FPGA. In sumulation we would load both (rom and test) and skip bootstrap. For that to work we need the gpio 8 to be pulled high during boot.
Maybe it's easier to pull all pin high (that's is what I'm going to do for verilator)?

hw/top_chip/dv/tb/tb.sv
```diff
@@ -33,6 +37,7 @@ module tb;
   clk_rst_if sys_clk_if(.clk(clk), .rst_n(rst_n));
   clk_rst_if peri_clk_if(.clk(peri_clk), .rst_n(peri_rst_n));
   uart_if uart_if();
+  pins_if #(NUM_GPIOS) gpio_if (.pins(gpio_pads));
```
marnovandermaas: A bit confusing that gpio_if is the name of pins_if, maybe gpio_pins_if?

hw/top_chip/dv/tb/tb.sv
marnovandermaas: Kinza showed me the pins_if.sv which already has a pull functionality. Hopefully we can use that and set all the pins to pull up.

marnovandermaas (commented):
Two minor comments.

hw/top_chip/dv/tb/tb.sv
KinzaQamar: Hi @engdoreis,

I've done it for every pin in this commit: fdbe886ef52406b5412b00a3c980fc0c5b14b8f6

I can pull out that commit along with 2b54119919a396f0b6809922c63334577aa642ba in a separate PR if your work is urgent?

hw/top_chip/dv/tb/tb.sv
martin-velay: This interface is actually used for other purposes and it comes from OT: https://github.com/lowRISC/opentitan/blob/master/hw/dv/sv/common_ifs/pins_if.sv

KinzaQamar: Hmmm, @marnovandermaas made this comment in front of me as he thought `gpio_if` is an interface specific to GPIO. But we are using a generic `pins_if`. @martin-velay I think your point is valid. But I am thinking about expanding GPIO chip level DV where we need to test `straps` and need [gpio_straps_if](https://github.com/lowRISC/opentitan/blob/master/hw/ip_templates/gpio/dv/interfaces/gpio_straps_if.sv.tpl) then we may end up with conflicts in instance names. Given that argument, I'll probably rename it as `gpio_io_if / gpio_pins_if`? Does that look reasonable to you?

martin-velay: Yes, it sounds good

hw/top_chip/dv/tb/tb.sv
```diff
@@ -190,13 +205,19 @@ module tb;
     `SIM_SRAM_IF.u_sw_test_status_if.sw_test_status_addr  = SW_DV_TEST_STATUS_ADDR;
     `SIM_SRAM_IF.u_sw_logger_if.sw_log_addr               = SW_DV_LOG_ADDR;
 
+    // Set the pull ups on GPIO pins. This doesn't mean anything for DV. Rather a feature added for
+    // SW. GPIO pin 8 is going to be used by the bootrom to decide the booting process. To support
+    // SW booting process in simulation, pin 8 needs to be pulled up during booting. But instead of
+    // initializing a single pin as pulled high, we'll do the same for evrey pin.
```
KinzaQamar: Hi @engdoreis 

Do you agree with this comment? Could you please double-check if this is true

KinzaQamar:
@engdoreis Could you take a look and approve it from SW side if everything looks okay?

@elliotb-lowrisc you've made a changes request which needs to get resolve to merge this PR. Could you take another look if all seems okay to you now?

@rswarbrick / @martin-velay I need an approval from DV side :))

Thanks everybody. I am hoping this is going to be the last round of reviews for this PR

sw/device/tests/gpio/smoketest.c
```diff
@@ -0,0 +1,36 @@
...
+
+    // Once the SW finishes driving the walking 1's pattern on the last GPIO pin, it should disable
+    // the output enables so that the GPIO pads can safely be driven as inputs
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0x0);
+
+    // Wait for the pattern of 1's on odd pins and 0's on even pins to be driven externally
+    while (DEV_READ(gpio + GPIO_REG_DATA_IN) != 0x55555555) {
+    }
+
```
elliotb-lowrisc, line 37: Is there some timeout mechanism elsewhere that will catch the case where this loops forever?

KinzaQamar: No, if this can't find `0x5555_5555` this will loop forever. Good point

engdoreis: In Opentitan tests would have a timeout, I don't know wether is controlled by dvsim or the tb. But I wuld argue that a general tb timeout would be the clenest way.
it can be tested adding a infinite loop somewhere, run the test and check that it would timeout.

engdoreis (approved):
Besides the timeout issue raised by Elliot, the SW side, LGTM.

sw/device/tests/gpio/smoketest.c
KinzaQamar: I am familiar with DVSim job timeout after 120, 180, 260 minutes. It is defined in the relevant `sim.hjson`.

sw/device/tests/gpio/smoketest.c
```diff
@@ -0,0 +1,45 @@
...
+    // Once the SW finishes driving the walking 1's pattern on the last GPIO pin, it should disable
+    // the output enables so that the GPIO pads can safely be driven as inputs
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0x0);
+
+    // Wait for the pattern of 1's on odd pins and 0's on even pins to be driven externally
+    while (DEV_READ(gpio + GPIO_REG_DATA_IN) != 0x55555555) {
+      // The expected pattern is driven externally on the inputs by toggling the wires. Since this
+      // occurs without delay in simulation, a timeout of 10 read requests is sufficient to detect a
+      // failure.
+      if (timeout == 0x10) {
```
KinzaQamar: @elliotb-lowrisc, I've implemented a timeout mechanism. Ideally, we'll never enter into the `while loop`. But I tested it by tweaking the expected pattern as: ` while (DEV_READ(gpio + GPIO_REG_DATA_IN) != 0x55510555) { blah }` and then it was able to fail.

elliotb-lowrisc (approved):
Thanks for adding the timeout

sw/device/tests/gpio/smoketest.c
```diff
@@ -0,0 +1,45 @@
...
+    // Once the SW finishes driving the walking 1's pattern on the last GPIO pin, it should disable
+    // the output enables so that the GPIO pads can safely be driven as inputs
+    DEV_WRITE(gpio + GPIO_REG_DIRECT_OE, 0x0);
+
+    // Wait for the pattern of 1's on odd pins and 0's on even pins to be driven externally.
+    while (DEV_READ(gpio + GPIO_REG_DATA_IN) != 0x55555555) {
+        // The expected pattern is driven externally on the inputs by toggling the wires. Since this
+        // occurs without delay in simulation, a timeout of 10 read requests is sufficient to detect
+        // a failure.
+        if (timeout == 0x10) {
```
KinzaQamar: This is not decimal 10, phew

sw/device/tests/gpio/smoketest.c
martin-velay: Yes for DV we can control timeouts via the HJSON `sim_cfg` files:
- DVSim:
`run_timeout_mins: 180`
But I don't see the default value, if any...

- Extra arg to the simulator:
By default it's set to 200ms. https://github.com/lowRISC/mocha/blob/3f74c05299b53047f5bd60833b0721badcba92ca/hw/top_chip/dv/test/top_chip_dv_base_test.sv#L11
And can be overridden from the HJSON file with:
`run_opts: ["+test_timeout_ns=120_000_000"]`

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_smoke_vseq.sv
```diff
@@ -17,5 +22,28 @@ endfunction : new
 task top_chip_dv_gpio_smoke_vseq::body();
   super.body();
   `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
-  // TODO
```
marnovandermaas: Can we squash these two commits so we don't introduce and remove the TODO?

hw/top_chip/dv/env/seq_lib/top_chip_dv_gpio_smoke_vseq.sv
```diff
@@ -25,6 +25,9 @@ task top_chip_dv_gpio_smoke_vseq::body();
 
   `uvm_info(`gfn, "Starting GPIOs outputs test", UVM_LOW)
 
+  // Disable pin pull-ups enabled to support SW booting process
+  cfg.gpio_vif.set_pullup_en('0);
+
```
marnovandermaas, line 30: Can we squash this commit with the other two where you change the gpio_smoke_vseq?

marnovandermaas (commented):
Just some minor comments on the commits, but otherwise this looks good to me.

KinzaQamar:
@marnovandermaas, I've squash the `vseq` related commits into one. Does that looks okay now?

marnovandermaas (approved):
This looks good to me! Thanks for the great work.

hw/top_chip/dv/top_chip_sim_cfg.hjson
martin-velay, line 257: I was about to press the merge button and noticed that the CHERI test for GPIO is missing here:
```suggestion
        "gpio_smoke",
        "gpio_smoke_cheri",
```

Without it the smoke regression won't run it. This command:
`dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i smoke`

I have tested to run these tests and both passed (even with coverage enabled).

martin-velay: Also add the CHERI version here please

martin-velay (approved):
Thanks for this PR, everything is looking good!

Regressions results:
`dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i smoke`

```
|  Stage  |  Name  | Tests                               |  Max Job Runtime  |  Simulated Time  |  Passing  |  Total  |  Pass Rate  |
|:-------:|:------:|:------------------------------------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
|         |        | test_framework_exception_test_cheri |      28.000s      |    1114.043us    |     1     |    1    |  100.00 %   |
|         |        | mailbox_smoke                       |      23.000s      |    265.015us     |     1     |    1    |  100.00 %   |
|         |        | rv_plic_smoke_cheri                 |      20.000s      |    712.428us     |     1     |    1    |  100.00 %   |
|         |        | gpio_smoke_cheri                    |      19.000s      |    620.460us     |     1     |    1    |  100.00 %   |
|         |        | rv_timer_smoke_cheri                |      23.000s      |    753.999us     |     1     |    1    |  100.00 %   |
|         |        | rv_timer_smoke                      |      24.000s      |    422.396us     |     1     |    1    |  100.00 %   |
|         |        | spi_device_smoke_cheri              |      30.000s      |    1091.393us    |     0     |    1    |   0.00 %    |
|         |        | uart_smoke                          |      30.000s      |    748.159us     |     1     |    1    |  100.00 %   |
|         |        | rom_ctrl_smoke                      |      88.000s      |    7919.169us    |     1     |    1    |  100.00 %   |
|         |        | rv_plic_smoke                       |      25.000s      |    368.954us     |     1     |    1    |  100.00 %   |
|         |        | axi_sram_smoke_cheri                |      24.000s      |    524.362us     |     1     |    1    |  100.00 %   |
|         |        | rom_ctrl_smoke_cheri                |      96.000s      |    9017.771us    |     1     |    1    |  100.00 %   |
|         |        | uart_smoke_cheri                    |      22.000s      |    746.157us     |     1     |    1    |  100.00 %   |
|         |        | rv_timer_irq                        |      34.000s      |    1796.329us    |     1     |    1    |  100.00 %   |
|         |        | spi_device_smoke                    |      24.000s      |    745.884us     |     0     |    1    |   0.00 %    |
|         |        | test_framework_exception_test       |      23.000s      |    731.095us     |     1     |    1    |  100.00 %   |
|         |        | rv_timer_irq_cheri                  |      37.000s      |    2194.132us    |     1     |    1    |  100.00 %   |
|         |        | gpio_smoke                          |      22.000s      |    286.347us     |     1     |    1    |  100.00 %   |
|         |        | spi_host_smoke                      |      19.000s      |    214.895us     |     1     |    1    |  100.00 %   |
|         |        | axi_sram_smoke                      |      8.000s       |    222.842us     |     1     |    1    |  100.00 %   |
|         |        | mailbox_smoke_cheri                 |      22.000s      |    580.866us     |     1     |    1    |  100.00 %   |
|         |        | spi_host_smoke_cheri                |      22.000s      |    522.319us     |     1     |    1    |  100.00 %   |
|         |        | **TOTAL**                           |                   |                  |    20     |   22    |   90.91 %   |
```

Both tests `gpio_smoke` and `gpio_smoke_cheri` are at 100%.

PR #385, opened 2026-03-26
Update Timer HAL to use new autogen interface

ziuziakowska:
Split off from https://github.com/lowRISC/mocha/pull/299.

### DIFF
--- a/sw/device/lib/hal/timer.c
+++ b/sw/device/lib/hal/timer.c
@@ -3,153 +3,132 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "hal/timer.h"
+#include "builtin.h"
 #include "hal/mmio.h"
+#include "hal/mocha.h"
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 
-void timer_disable(timer_t timer)
-{
-    DEV_WRITE(timer + TIMER_CTRL_REG, 0x0);
-}
-
-void timer_enable(timer_t timer)
-{
-    DEV_WRITE(timer + TIMER_CTRL_REG, 0x1);
-}
-
-bool timer_get_enable(timer_t timer)
-{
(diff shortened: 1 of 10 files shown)


sw/device/lib/hal/timer.c
```diff
@@ -4,152 +4,116 @@
...
-uint32_t timer_get_value_upper(timer_t timer)
+bool timer_interrupt_pending(timer_t timer)
 {
-    return DEV_READ(timer + TIMER_V_UPPER0_REG);
+    timer_intr_state0 intr_state = VOLATILE_READ(timer->intr_state0);
+    return intr_state.is;
 }
 
-uint64_t timer_get_value(timer_t timer)
+void timer_enable(timer_t timer)
```
engdoreis: Nit. I would prefer `timer_set_enabled(timer_t timer, bool enable)`

sw/device/lib/hal/timer.c
```diff
@@ -3,153 +3,138 @@
...
-void timer_busy_sleep(timer_t timer, uint64_t duration_steps)
+void timer_busy_sleep_us(timer_t timer, uint64_t us)
 {
-    timer_set_compare(timer, timer_get_value(timer) + duration_steps);
-    timer_clear_interrupt(timer);
-
-    // Poll for interrupt
-    while (!timer_has_interrupt(timer)) {
+    timer_schedule_in_us(timer, us);
+    while (!timer_interrupt_pending(timer)) {
```
engdoreis: I guess that a 'wfi' could be added here?

engdoreis (approved):
LGTM.

sw/device/lib/hal/timer.c
ziuziakowska: I think I will leave this as a busy loop for now - for a `wfi` we would have to set and restore the `mie.mtie` bit as `wfi` honours individual interrupt enables, but it means if interrupts are globally enabled then this busy sleep would be pre-empt when the timer has expired which might be unexpected.

PR #392, opened 2026-03-31
DRAM loading in Verilator

### LINKED ISSUE #377: DRAM in Verilator simulation
When trying to load DRAM with a binary using `-E`, the following error occurs:

```
ERROR: Failed to load ELF file at `build/sw/device/tests/test_framework_test_vanilla_rom': No memory region is registered that contains the address 0x80000000 (the base address of segment 0).
```

ziuziakowska:
Adds the DRAM as a memory area that can be loaded on the command line, so that we can skip the bootloader in Verilator in the future.

The memory can be loaded into using multiple `-E` flags to load binaries into the corresponding memory areas, ~~or using the `-r` and `-m` flags to explicitly load into ROM/SRAM or DRAM.~~

~~We currently don't use this functionality as we still need a Verilator bypass in the boot ROM before we can run all tests from DRAM on all environments.~~

Added a test now, needed quite a bit of machinery to get it going (needed the ability to load multiple ELF's with the test runner, compile and link for DRAM which we do not do yet) but that should be cleaned up once we can load all our tests into DRAM in addition to just SRAM.

Closes https://github.com/lowRISC/mocha/issues/377.

### DIFF
--- a/hw/top_chip/dv/verilator/top_chip_verilator.cc
+++ b/hw/top_chip/dv/verilator/top_chip_verilator.cc
@@ -12,7 +12,8 @@
 
 class MochaSim {
  public:
-  MochaSim(const char *ram_hier_path, int ram_size_words);
+  MochaSim(const char *sram_hier_path, int sram_size_words,
+           const char *dram_hier_path, int dram_size_words);
   virtual ~MochaSim() {}
   virtual int Main(int argc, char **argv);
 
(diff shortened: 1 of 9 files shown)


engdoreis (commented):
LGTM, but I think that Marno is the best to review this.
Can you please add a verilator DRAM smoketest to make sure this will not break in follow up PRs.

marnovandermaas (commented):
It would be good to have a test in CI for making sure that this works. Is that something you can add before we merge this PR?

marnovandermaas (commented):
This is great, is there a way we can test this change before merging? Preferably in CI. That's the only thing stopping me from approving this PR right now.

engdoreis (commented):
I'm sorry for not being clear when I asked for a test. I had in mind a test similar in structure to [this one](https://github.com/lowRISC/sonata-system/blob/main/sw/legacy/test/memory_test.cc). The one you implemented in this PR is not wrong, but since it is basically the bootroom in the near feature, so it won't add extra coverage.

engdoreis (approved):

marnovandermaas (approved):
I'm happy for this to be merged.

PR #398, opened 2026-04-03
util: FPGA runner: rework to use asyncio

ziuziakowska:
Was https://github.com/lowRISC/mocha/pull/366 but that got auto-closed for some reason.
Split off from: https://github.com/lowRISC/mocha/pull/299

This reworks the FPGA runner to use asyncio to read from the UART while the flash is being written. This was causing some problems for me locally when adding some more debugging output, as in the time between finishing flashing and opening the UART file descriptor, the receive buffer would be overflowed and the test status would be dropped.

### DIFF
--- a/util/fpga_runner.py
+++ b/util/fpga_runner.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
-import subprocess
+import asyncio
 import sys
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


util/fpga_runner.py
```diff
@@ -16,49 +16,44 @@
...
 
+RUNNER: str = Path(__file__).name
 
-def load_fpga_test(test: Path, timeout: int = TIMEOUT) -> None:
+
+async def load_fpga_test(test: Path) -> None:
     command = ["openFPGALoader", "--spi", "--offset", str(BOOT_ROM_OFFSET), "--write-flash"]
-    command.append(test.with_suffix(".bin"))
+    command.append(str(test.with_suffix(".bin")))
+    # TODO: This is a workaround to send a reset and start the test, should be removed when we
```
engdoreis: Please keep the TODO comment before the second time the command is run, since the second run is the workaround.

engdoreis (approved):
LGTM, please address my comment before merging.

PR #406, opened 2026-04-08
[chip, gpio, dv] Renaming Verilator GPIO's smoketest.c to reg_access_test.c

KinzaQamar:
Merge this PR first as a dependency on #382 

~~The motivation of this PR comes from #382 which is adding a chip level GPIO test for Xcelium. That test fails on verilator. The reason being different tops used for both of the tools. For Xcelium, an interface is connecting the UVM to the GPIO. The UVM have the control to drive GPIOs inputs. In #382, the SW waits for a specific pattern driven by UVM to appear in `data_in` reg. If we run the exact test with the verilator top, which is using GPIO DPI, then the abilty to drive the inputs depends on a host SW infrastructure which is currently missing.~~

~~With the above context, the best way forward is to separate the GPIO smoketest for both of the simulators for now to make the CI happy.~~

Renaming the gpio/smoketest.c to gpio/reg_access_test.c. Reason being that #382 is adding another test which will check if every outputs and inputs can be driven in high/low combinations. That is a much refined definition of a smoke test. The second reason of renaming gpio/smoketest.c to gpio/reg_access_test.c is that the test added in #382 is incompatible for Verilator and FPGA tops. This also requires to ask the build system to only make images of GPIO tests for their compatible tops. So, the better approach is to rename the test after their features and add additional options to the build system in order to make their images for different platforms.


sw/device/tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ set(LIBS hal runtime startup test_framework)
 
 mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME test_framework_exception_test SOURCES test_framework/exception.c LIBRARIES ${LIBS})
-mocha_add_test(NAME gpio_smoketest SOURCES gpio/smoketest.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME gpio_smoketest_verilator SOURCES gpio/smoketest_verilator.c LIBRARIES ${LIBS} FPGA)
```
martin-velay: I am unsure about this modification, as the `mocha_add_test` will call this other macro [`mocha_add_verilator_test`](https://github.com/lowRISC/mocha/blob/0e4795c9d0dd0ba2779d0257063c15acb9d4b095/sw/cmake/tests.cmake#L35-L40) and thus create a test with `_sim_verilator` suffix.

@engdoreis, how to handle this kind of situation?

KinzaQamar: Hi Martin,

My idea was to create another gpio_smoketest_xcelium and ask the build system to stop making FPGA and Verialtor images if the target name contains xcelium in it. So, we don't get `gpio_smoketest_xcelium_sim_verilator`. I am not sure if there is a better way to doing thing but that is making the CI happy on #382

engdoreis: In my understanding this test can run in any platform (FPGA, VERILATOR and DV_SIM) right? So, we should not name it after verilator.

I suggest keeping this tests exactly the same. Then ~~copy it to another file, extend it with the features that's~~ create another test only supported by DV_SIM (xcelium), then name it `<feature_tested>.c`, then adapt cmake to mark this new test as not runnable in verilator.

Does that make sense?

martin-velay (approved):
I have just changed the title of this PR to precise that this test is only for Verilator. Can you also update the description to make sure it's clear?

KinzaQamar:
Thanks @martin-velay. I've added the description and pasted it in my commit message as well

engdoreis (approved):

PR #413, opened 2026-04-08
Enable bootrom bootstraping control using GPIOs

engdoreis:


sw/device/bootrom/bootrom.c
```diff
@@ -28,8 +30,12 @@ int main(void)
     uart_init(console);
     uprintf(console, "\nBoot ROM!\n");
 
-    // Spin polling the spi_dev and processng incoming data until a reset command is received.
-    spi_boot_strap(console);
+    gpio_t gpio = mocha_system_gpio();
+    if (bootstrap_requested(gpio)) {
+        uprintf(console, "Entering spi bootstrap\n");
+        // Spin polling the spi_dev and processng incoming data until a reset command is received.
```
ziuziakowska: Typo: processing

util/fpga_runner.py
```diff
@@ -15,31 +15,59 @@
...
+        print(line, end="")
+        if not line or "Entering spi bootstrap" not in line:
+            continue
+
+        await set_pin(0, 1)
+        return 0
 
-async def load_fpga_test(test: Path) -> None:
-    command = ["openFPGALoader", "--spi", "--offset", str(BOOT_ROM_OFFSET), "--write-flash"]
+    print(f"[{RUNNER}]  Bootstrap timeout")
```
ziuziakowska: Nit: Extra space between `]` and `B`.

util/fpga_runner.py
```diff
@@ -15,31 +15,59 @@
...
+async def bootstrap(uart) -> None:
+    """Pull down the bootstrap pin, rset the core then release the pin 
+    """
+    await set_pin(0, 0)
+    await reset_core()
+    start = time.time()
+    while time.time() - start < TIMEOUT:
+        line = uart.readline().decode("utf-8", errors="ignore")
+        print(line, end="")
+        if not line or "Entering spi bootstrap" not in line:
```
ziuziakowska: I think this string should go above with the other constants as it is special, perhaps called `SPI_BOOTSTRAP_MAGIC`? Also I think it should be written as `SPI Bootstrap`.

util/fpga_runner.py
ziuziakowska, line 70: Forgot this comment!

util/fpga_runner.py
```diff
@@ -15,31 +15,59 @@
...
+
+async def reset_core() -> None:
+    """Pull the reset pin down for 100ms.
+    """
+    await set_pin(2, 0)
+    time.sleep(0.1)
+    await set_pin(2, 1)
+
+async def bootstrap(uart) -> None:
+    """Pull down the bootstrap pin, rset the core then release the pin
```
ziuziakowska:
```suggestion
    """Pull down the bootstrap pin, reset the core, then release the pin."""

File: util/fpga_runner.py [hunk 5]
```diff
@@ -15,31 +15,59 @@
...
 
+async def set_pin(pin:int, val:int) -> None:
+    command = ["ftditool", "gpio-write", str(pin), str(val), '--ftdi', FTDI_DEVICE_DESC]
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
+        print(f"[{RUNNER}] gpio-write command exited with non-zero exit code {p.returncode}")
+        sys.exit(1)
+
+async def reset_core() -> None:
+    """Pull the reset pin down for 100ms.
```
ziuziakowska: I'm not sure if the linter has an opinion on this, but I would prefer newlines after each `"""`.

util/fpga_runner.py
```diff
@@ -55,7 +83,6 @@ async def poll_uart(uart: serial.Serial) -> bool:
     print(f"[{RUNNER}] Test timeout")
     return False
 
-
```
ziuziakowska, line 58: If I recall correctly, `ruff` wants two empty lines between functions

util/fpga_runner.py
```diff
@@ -15,31 +15,59 @@
...
+            continue
+
+        await set_pin(0, 1)
+        return 0
 
-async def load_fpga_test(test: Path) -> None:
-    command = ["openFPGALoader", "--spi", "--offset", str(BOOT_ROM_OFFSET), "--write-flash"]
+    print(f"[{RUNNER}]  Bootstrap timeout")
+    await set_pin(0, 1)
+    return 1
```
ziuziakowska: The returns here don't match the type hint in the signature

ziuziakowska (commented):
Initial review, thanks for the important work! Just some nits.

hw/top_chip/rtl/chip_mocha_genesys2.sv
```diff
@@ -119,7 +120,7 @@ module chip_mocha_genesys2 #(
     .ResetValue ('0)
   ) u_mig_rst_sync_cfg (
     .clk_i  (clk_cfg),
-    .rst_ni (ext_rst_ni & fpga_rst_n_sync_cfg),
+    .rst_ni (ext_rst_ni & fpga_rst_n_sync_cfg & ftdi_rst_ni),
```
AlexJones0: Really minor nit:

```suggestion
    .rst_ni (ext_rst_ni & ftdi_rst_ni & fpga_rst_n_sync_cfg),
```

For the RTL changes to the external reset synchronizer(s), can you change the expression like the above to put the external resets first and the initial reset logic afterwards?

hw/top_chip/data/pins_genesys2.xdc
```diff
@@ -19,6 +20,11 @@ set_property -dict { PACKAGE_PIN N19   IOSTANDARD LVCMOS18 } [get_ports { gpio_i
 set_property -dict { PACKAGE_PIN P19   IOSTANDARD LVCMOS18 } [get_ports { gpio_i[5] }]; # SW5 (VADJ)
 set_property -dict { PACKAGE_PIN P26   IOSTANDARD LVCMOS33 } [get_ports { gpio_i[6] }]; # SW6 (VCC3V3)
 set_property -dict { PACKAGE_PIN P27   IOSTANDARD LVCMOS33 } [get_ports { gpio_i[7] }]; # SW7 (VCC3V3)
+
+# Bootstrap pin, should be pulled down during boot to enter bootstrap mode.
+set_property -dict { PACKAGE_PIN AB29  IOSTANDARD LVCMOS33 } [get_ports { gpio_i[8] }];
+set_property PULLTYPE PULLUP [get_ports gpio_i[8]];
```
AlexJones0: I think we should go with this formatting to match the other pins in this file (e.g. I2C, SPI), and to better display a single source of truth about the pin properties.
```suggestion
set_property -dict { PACKAGE_PIN AB29  IOSTANDARD LVCMOS33  PULLTYPE PULLUP } [get_ports { gpio_i[8] }];
```

util/fpga_runner.py
```diff
@@ -37,23 +38,24 @@ async def load_fpga_test(test: Path) -> None:
 async def run_fpga_test(tty: str, test: Path) -> bool:
     with serial.Serial(tty, BAUD_RATE, timeout=0) as uart:
         load = asyncio.create_task(load_fpga_test(test))
-        poll = asyncio.create_task(poll_uart(uart))
-        success = await poll
-        await load
-        return success
+        poll = asyncio.create_task(poll_uart_checking_for(uart, r"TEST RESULT: (PASSED|FAILED)"))
+        result = await poll
+        return result and "PASSED" in result
```
AlexJones0, line 95: Nit: the `await load` is dropped in this commit and only fixed in the subsequent commit, which may break bisecting flows.

Regardless, I missed when I originally looked at this code, but a more idiomatic and stylistically clearer version would be:

```py
_, result = await asyncio.gather(load_fpga_test(test), poll_uart(uart))
return result and "PASSED" in result
```

util/fpga_runner.py
```diff
@@ -15,45 +16,72 @@
...
-        await load
-        return success
+        await load_fpga_test(test, uart)
+        poll = asyncio.create_task(poll_uart_checking_for(uart, r"TEST RESULT: (PASSED|FAILED)"))
+        result = await poll
+        return result and "PASSED" in result
 
 
-async def poll_uart(uart: serial.Serial) -> bool:
+async def poll_uart_checking_for(uart: serial.Serial, pattern: str) -> None | str:
```
AlexJones0: Style nit:

```suggestion
async def poll_uart_checking_for(uart: serial.Serial, pattern: str) -> str | None:
```

util/fpga_runner.py
```diff
@@ -16,28 +16,54 @@
...
 
 RUNNER: str = Path(__file__).name
 
 
-async def load_fpga_test(test: Path) -> None:
-    command = ["openFPGALoader", "--spi", "--offset", str(BOOT_ROM_OFFSET), "--write-flash"]
+async def set_pin(pin: int, val: int) -> None:
+    command = ["ftditool", "gpio-write", str(pin), str(val), "--ftdi", FTDI_DEVICE_DESC]
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
```
AlexJones0: I think this probably won't happen, but it could be a good idea to add a timeout to this in case ftditool hangs for some reason. Something like:
```py
    try:
        async with asyncio.timeout_at(FTDITOOL_WAIT_TIME):
            if await p.wait() != 0:
                print(f"[{RUNNER}] gpio-write command exited with non-zero exit code {p.returncode}")
                sys.exit(1)
    except TimeoutError:
        print(f"[{RUNNER}] gpio-write timed out after {FTDITOOL_WAIT_TIME} seconds.")
        await p.terminate()  # or maybe p.kill()
        sys.exit(1)
```

util/fpga_runner.py
```diff
@@ -16,28 +16,54 @@
...
+        sys.exit(1)
+
+
+async def reset_core() -> None:
+    """
+    Pull the reset pin down for 100ms.
+    """
+    await set_pin(2, 0)
+    time.sleep(0.1)
+    await set_pin(2, 1)
```
AlexJones0, line 56: Seems a little magic from reading, can you name the `FTDI_RESET_PIN=2` constant (or similar) at least? Probably we need similar for the `FTDI_BOOTSTRAP_PIN=0` below as well?

Also, for async prefer:
```py
await asyncio.sleep(0.1)
```
over
```py
time.sleep(0.1)
```

sw/device/bootrom/bootrom.c
```diff
@@ -144,19 +148,41 @@ void led_init(gpio_t gpio)
...
     int next_led = current_led + (going_up ? 1 : -1);
     bool toggle = (next_led >= num_leds || next_led < 0);
     current_led = toggle ? current_led : next_led;
     going_up ^= toggle;
 }
 
+bool bootstrap_requested(struct context *ctx)
+{
+    enum { bootstrap_pin = 8, debaunce_us = 20 * 1000 };
+    timer_schedule_in_us(ctx->timer, debaunce_us);
```
AlexJones0: Typo:
```suggestion
    enum { bootstrap_pin = 8, debounce_us = 20 * 1000 };
    timer_schedule_in_us(ctx->timer, debounce_us);
```

sw/device/bootrom/bootrom.c
```diff
@@ -11,41 +11,49 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+struct context {
```
AlexJones0: Nit: rename `context` to `boot_context` to be more specific?

sw/device/lib/hal/timer.c
```diff
@@ -132,3 +132,8 @@ void timer_busy_sleep_us(timer_t timer, uint64_t us)
     while (!timer_interrupt_pending(timer)) {
     }
 }
+
+uint64_t timer_value_read_us(timer_t timer)
+{
+    return timer_value_read(timer) / timer_ticks_per_us(timer);
```
AlexJones0, line 138: Question: should we be concerned about potential imprecision due to integer division on the `timer_ticks_per_us` here with prescalers set which will grow for larger timer values? Or it this not a concern?

AlexJones0 (commented):
Looks like a big improvement to me and gets rid of a lot of the hacks that we had! I have a collection of minor nitpicks which can be quickly fixed, but the overall changes LGTM.

sw/device/lib/hal/timer.c
engdoreis: I don't think this is a concern for the pourpose of this function.

util/fpga_runner.py
```diff
@@ -15,45 +16,79 @@
 BOOT_ROM_OFFSET: int = 0x4000
 BAUD_RATE: int = 1_000_000
 TIMEOUT: int = 60
+FTDITOOL_WAIT_TIME: int  = 60
```
AlexJones0:
```suggestion
FTDITOOL_WAIT_TIME: int = 60
```

flake.nix
```diff
@@ -24,7 +24,8 @@
       inputs.uv2nix.follows = "uv2nix";
     };
     ftditool = {
-      url = "github:lowRISC/ftditool?ref=v0.1.0";
+      # url = "github:lowRISC/ftditool?ref=v0.1.0";
```
AlexJones0: I think you want to squash/fixup this second commit (judging by this name)? Or maybe you still haven't finished the changes here, given that it's pinning a pull request at the moment.

AlexJones0 (approved):
LGTM, thanks @engdoreis.

AlexJones0 (approved):

marnovandermaas (approved):

Issue #303, opened 2026-02-19
[design] Integrate debug module into Mocha top


marnovandermaas:
Based on https://github.com/pulp-platform/riscv-dbg/tree/master

Part of this task includes:
- Integrate the RTL like we did for SPI device in this PR: https://github.com/lowRISC/mocha/pull/259
- Use the correct memory map as declared here: https://github.com/lowRISC/mocha?tab=readme-ov-file#memory-map
- The debug module must be a manager on the AXI crossbar.
- Add a section to our readme to say how we can use the debug module with the Genesys 2 board.

marnovandermaas:
Conversation with Jon captured here:
> Hey Marno!  I'm looking at the u_clk_mgr; is there a reset that you all had envisiged being used for the ndmreset_n?  The debug module spits out a reset that CVA6 and things connected to it are generally reset by.  I guess this should go through the new reset manager in some way?

Hi Jon, I tend to look at OpenTitan to see what is done there. It is connected here in the debug module: https://github.com/lowRISC/opentitan/blob/222916aeaa415d16f6aba9201a367f3d5447f78c/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L2417 and is fed into the power manager: https://github.com/lowRISC/opentitan/blob/222916aeaa415d16f6aba9201a367f3d5447f78c/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L1910 Let me know if that helps.

> Ok, I'm a bit stuck in the mud.  In Earl Grey, there are _lc_ resets and _sys_ resets.  The _lc_  ones reset ~everything except the debug module, and _sys_  resets everything. This is documented [here](https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rstmgr/doc/theory_of_operation.html?utm_source=chatgpt.com), and reflected in the source that you linked me to.
> In the lowrisc/main branch of mocha (as of last week), in [top_chip_system.sv](http://top_chip_system.sv/), instead we have some _io_  signals and _main_  signals.  Deeper in the hierarchy, we do still have _lc_  and _sys_  signals, but when I chase them through, I can't convince myself that they correspond in some clear way to _io_  and _main_ .
[...]

PR #436, opened 2026-04-10
Debug Module ongoing integration into Mocha

alexandrecapltd:
This PR captures ongoing work to integrate the riscv-dbg module into mocha, following up on #303.

Currently,
- The pulp riscv-dbg module has been vendored 
- The vendored lowrisc_ip now include the `jtagdpi` and `tcp_server` ips
- A `jtagdpi` instance has been added to `hw/top_chip/dv/verilator/top_chip_verilator.sv` and connected the `top_chip_system` module jtag pins.
- `hw/top_chip/rtl/top_chip_system.sv` now has a debug module instance connected through converter modules (`dmi_jtag` for the debug module interface, `axi_adapter` for the master interface and `axi_to_mem` for the slave interface). The jtag pins are exported. The `debug_req_irq` is connected to the CVA6 core, and the reset coming out of the debug module is in use (`ndmreset_n = (~ndmreset) & rst_ni`).
- In `hw/top_chip/rtl/top_pkg.sv`, the SoC configuration accounts for the added debug module.

This now builds for me, so I open this PR for others to be able to take a look.
I lifted the earlgrey openocd configuration from the opentitan repository. When simulating and trying to attach, the generated traces show traffic on the jtag interface, as well as a debug irq eventually being triggered. The CVA6 core which was emitting traffic on its noc_req signals goes silent, and openocd reports that the core failed to halt.

Things to investigate:
- Should some uses of ndmreset be using rst_ni?
- Is the specific version of the riscv-dbg module picked for vendoring the desired one?
- Is the set of patches (lifted from the opentitan repository ) currently applied the desired set?


hw/top_chip/rtl/top_chip_system.sv
marnovandermaas, line 5: Nit: introduce an extra newline

marnovandermaas, line 196: Please use manager and subordinate in sources that are not vendored in.

marnovandermaas, line 221: This'll need to be redone because I've integrated the reset manager. You should be able to pass this into the power manager like we do in Earl Grey: https://github.com/lowRISC/opentitan/blob/a1571d0a9873b15b218405ef368a0f89fecd06e9/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L1886

marnovandermaas, line 307: Nit: remove commented code.

hw/top_chip/rtl/top_pkg.sv
marnovandermaas: Debug module is at 0x2000_0000, see https://github.com/lowRISC/mocha/blob/main/doc/ref/arch.md

marnovandermaas, line 52: Is this because of the extra host?

hw/vendor/pulp_riscv_dbg.core
marnovandermaas, line 29: Very nice, I should do this for the other vendored core files as well.

hw/vendor/pulp_riscv_dbg.vendor.hjson
marnovandermaas, line 5: This vendor file should also be added to artefacts.py so we can check it in CI.

marnovandermaas (commented):
Good initial go at this. It would make it easier to review if you split this PR up into some separate commits. It would also be good if you could add some comments on how to test the code changes.

I've added some initial comments. It's probably worth rebasing to include the power and reset managers.

marnovandermaas:
In terms of which debug module to pull in, I think this should be the same as the one you are using for debugging in the CapLtd CVA6 repository.

PR #441, opened 2026-04-15
[Bootrom]  Enable SPI 4 bytes addressing

engdoreis:
This will support running tests in DRAM and SRAM.

### DIFF
--- a/sw/device/bootrom/bootrom.c
+++ b/sw/device/bootrom/bootrom.c
@@ -13,6 +13,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#define MAJOR "00"
+#define MINOR "01"
+#define PATCH "00"
 
 const uintptr_t boot_slots[] = { 0x10004000, 0x80000000 };
 struct boot_context {
@@ -44,7 +47,7 @@ int main(void)
     };
 
     uart_init(boot_ctx.console);
-    uprintf(boot_ctx.console, "\nBoot ROM!\n");
+    uprintf(boot_ctx.console, "\nBoot ROM: v%s.%s.%s\n", MAJOR, MINOR, PATCH);
 
     timer_init(boot_ctx.timer);
     timer_enable_write(boot_ctx.timer, true);
--- a/sw/device/bootrom/bootrom.c
+++ b/sw/device/bootrom/bootrom.c
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "boot/trap.h"
+#include "builtin.h"
+#include "constants.h"
 #include "hal/gpio.h"
 #include "hal/mocha.h"
 #include "hal/spi_device.h"
@@ -11,6 +13,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+
+const uintptr_t boot_slots[] = { 0x10004000, 0x80000000 };
 struct boot_context {
     uart_t console;
     gpio_t gpio;
@@ -27,6 +31,7 @@ static void boot(uintptr_t addr);
 static void led_init(gpio_t gpio);
 static void led_animation_run(struct boot_context *ctx);
 static bool bootstrap_requested(struct boot_context *ctx);
+static bool get_boot_addr(uint32_t *addr);
 
 
 // TODO: Add support to cheri mode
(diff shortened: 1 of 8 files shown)


util/fpga_runner.py
```diff
@@ -68,6 +68,15 @@ async def bootstrap(uart: serial.Serial) -> bool:
     return result is not None
 
 
+def get_load_addr(elf: Path) -> int:
+    with elf.open("rb") as f:
+        elf = ELFFile(f)
+        load_addrs = [seg["p_paddr"] for seg in elf.iter_segments() if seg["p_type"] == "PT_LOAD"]
+
+        first_address = min(load_addrs)
+        return first_address
```
AlexJones0: Nit: some more robust error handling with appropriate logs would be good here.

Specifically, something like:
```py
try:
    with elf.open("rb") as f:
        elf = ELFFile(f)
        load_addrs = [seg["p_paddr"] for seg in elf.iter_segments() if seg["p_type"] == "PT_LOAD"]
        if not load_addrs:
            print(f"[{RUNNER}] no `PT_LOAD` segments found in elf {elf}")
            sys.exit(1)

        first_address = min(load_addrs)
        return first_address
except OSError as e:
    print(f"[{RUNNER}] error opening elf {elf}: {e}")
    sys.exit(1)

File: sw/device/lib/hal/spi_device.c [hunk 2]
```diff
@@ -214,6 +208,20 @@ uint32_t spi_device_cmd_info_get(spi_device_t spi_device, uint32_t offset)
...
 
+void spi_device_cmd_info_4b_enable_set_raw(spi_device_t spi_device, uint32_t data)
+{
+    DEV_WRITE(spi_device + SPI_DEVICE_CMD_INFO_EN4B_REG, data);
+}
+
+void spi_device_cmd_info_4b_enable_set(spi_device_t spi_device, uint8_t opcode)
+{
+    uint32_t reg = 0;
+    reg = reg | (opcode << SPI_DEVICE_CMD_OPCODE);
```
AlexJones0, line 219: Really minor nit (not necessary to fix in this PR). Subjectively I prefer `reg |= (opcode << SPI_DEVICE_CMD_OPCODE);` since it removes the needless duplication of `reg` but I see this is in many places in this file. Maybe we could change it in a different PR?

sw/device/lib/hal/spi_device.c
```diff
@@ -575,25 +583,29 @@ void spi_device_init(spi_device_t spi_device)
     spi_device_flash_status_set(spi_device, 0);
 
     // Configure commands
-    spi_device_cmd_info_set(spi_device, SPI_DEVICE_CMD_INFO_0_REG, SPI_DEVICE_OPCODE_READ_STATUS,
-                            false, 0, false);
+    spi_device_cmd_info_set(spi_device, SPI_DEVICE_CMD_INFO_0_REG, SPI_DEVICE_OPCODE_READ_STATUS, 0,
```
AlexJones0, line 586: Repeated a few different times - even for `0`, I think we should use the constant name:


```suggestion
    spi_device_cmd_info_set(spi_device, SPI_DEVICE_CMD_INFO_0_REG, SPI_DEVICE_OPCODE_READ_STATUS, SPI_DEVICE_CMD_ADDR_MODE_ADDR_DISABLED,
```

sw/device/bootrom/bootrom.c
```diff
@@ -64,6 +72,22 @@ void boot(uintptr_t addr)
     reset();
 }
 
+bool get_boot_addr(struct boot_context *ctx, uint32_t *addr)
+{
+    for (size_t i = 0; i < sizeof(boot_slots) / sizeof(boot_slots[0]); i++) {
```
AlexJones0: Nit: maybe we should introduce an `ARRAYSIZE` macro to make this more readable?

sw/device/bootrom/bootrom.c
```diff
@@ -64,6 +72,22 @@ void boot(uintptr_t addr)
...
+        uintptr_t slot = boot_slots[i];
+        uprintf(ctx->console, "checking: 0x%x\n", (uint32_t)slot);
+        if (*((volatile uint32_t *)slot) == 0xbebacafe) {
+            slot += sizeof(uint32_t);
+            *addr = *((volatile uint32_t *)slot);
```
AlexJones0: Nit: do we not have any macros that can be used for these volatile accesses?

sw/device/bootrom/bootrom.c
```diff
@@ -11,6 +11,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#define MAJOR   00
+#define MINOR   01
+#define BUG_FIX 00
```
AlexJones0: Nit: maybe rename `BUG_FIX` to `PATCH` to fit common SemVer convention?

AlexJones0 (approved):
LGTM, just a few nits.

sw/device/lib/hal/spi_device.c
engdoreis: I agree, but I don't want to change the spi_device.c because it will replaced by the soon.

sw/device/lib/hal/spi_device.c
engdoreis: I agree, but I don't want to change the spi_device.c because it will replaced by the soon.

Issue #411, opened 2026-04-08
[fpga] Genesys2 top-level port names don't match schematic names


elliotb-lowrisc:
Before we go too much further, I think it is worth considering whether we should switch to naming the ports in chip_mocha_genesys2.sv and pins_genesys2.xdc after the corresponding net name on the Genesys2 schematic rather than the current mixed approach (for which I'm partly to blame).

marnovandermaas:
I think this makes sense to me, this then means the chip_mocha_genesys2.sv is in charge of mapping those names onto the relevant signals in top_chip_system.sv

PR #446, opened 2026-04-17
[spi_host] Low-level SD card support

### LINKED ISSUE #402: [hw] SPI host SD card integration
This should use the SPI device DPI from Sonata: https://github.com/lowRISC/sonata-system/tree/main/dv/dpi/spidevicedpi
And also the Genesys 2 SD card slot.

### LINKED ISSUE #534: Merge SD card HW/SW changes to Mocha
Go through the review process and associated code revisions for the SD card access HW and SW

elliotb-lowrisc:
Replace the initial SPI Host loopback connections with connections to the microSD card slot for Genesys2 and an SD card model for Verilator. This includes allocating a GPI pin for card presence detection and a GPO pin for SD card reset (power supply switching).

Also output the SPI Host signals on an otherwise unused Genesys2 PMOD header for easier debugging using a logic analyser.

Add SD card and minimal filesystem helpers for SPI Host, and some tests. These have been ported from sonata-system, which required changes for the different spi_host hardware and translating from C++ to C.

Only a subset of the tests will run in Verilator unless an SD card image is provided.


sw/device/lib/runtime/filesys_utils.h
```diff
@@ -0,0 +1,195 @@
...
+typedef enum fs_utils_dir_flags {
+    DirFlag_Raw = 1u,
+    DirFlag_IncludeDeleted = 2u,
+    DirFlag_IncludeHidden = 4u,
+
+    DirFlags_Default = 0u,
+} fs_utils_dir_flags_t;
+
+// Directory entry type flags; this just makes the most common types of entries more accessible.
+typedef enum fs_utils_dir_entry_flags {
```
engdoreis: Add this flag to avoid clang-tidy error: https://github.com/lowRISC/mocha/blob/57ab0e046f98852a6fdfb0534ddc5d83719bb50c/sw/device/lib/hal/mocha_irq.h#L12

elliotb-lowrisc: I've tried adding the tag, but it still seems to complain (see latest CI run)

elliotb-lowrisc: Ah, it seems I also needed to tweak the variable usage into something the tool is more familiar with for flags.

elliotb-lowrisc:
Reduced the level of UART logging when `logging = false` to avoid timing-out in the Verilator test suite

hw/top_chip/data/pins_genesys2.xdc
```diff
@@ -71,3 +76,16 @@ set_property -dict { PACKAGE_PIN AJ11  IOSTANDARD LVCMOS15 } [get_ports { eth_tx
 set_property -dict { PACKAGE_PIN AK10  IOSTANDARD LVCMOS15 } [get_ports { eth_tx_d[3]  }];
 set_property -dict { PACKAGE_PIN AF12  IOSTANDARD LVCMOS15 } [get_ports { eth_mdc      }];
 set_property -dict { PACKAGE_PIN AG12  IOSTANDARD LVCMOS15 } [get_ports { eth_mdio     }];
+
+## SPI Host (MicroSD card slot)
+set_property -dict { PACKAGE_PIN R28   IOSTANDARD LVCMOS33 } [get_ports { spi_host_sck_o }]; # SD_SCLK
+set_property -dict { PACKAGE_PIN R26   IOSTANDARD LVCMOS33 } [get_ports { spi_host_sd_i  }]; # SD_DAT0
+# set_property -dict { PACKAGE_PIN R30   IOSTANDARD LVCMOS33 } [get_ports { microsd_dat1 }]; # SD_DAT1 unused in SPI bus mode
```
engdoreis: Any reason to keep the commented line?

sw/device/lib/hal/uart.c
```diff
@@ -106,3 +107,27 @@ void uart_puts(uart_t uart, const char *str)
         uart_putchar(uart, *str++);
     }
 }
+
+// Dump out a sequence of bytes as hexadecimal and ASCII text.
+void uart_dump_bytes(uart_t uart, const uint8_t buf[], size_t len)
+{
+    for (size_t off = 0u; off < len; ++off) {
+        uart_putchar(uart, '0' + (buf[off] >> 4));
```
engdoreis: Please use the printf functions to implement the dump:
https://github.com/lowRISC/mocha/blob/main/sw/device/lib/runtime/print.h

sw/device/lib/runtime/filesys_utils.c
```diff
@@ -0,0 +1,852 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "runtime/filesys_utils.h"
+#include "runtime/print.h"
+#include "runtime/sdcard.h"
+// #include <assert.h>
```
engdoreis: drop commented line.

sw/device/lib/runtime/filesys_utils.c
```diff
@@ -0,0 +1,852 @@
...
+static size_t object_read(fs_utils_state_t *fs, fs_utils_obj_state_t *obj, uint8_t buf[],
+                          size_t len);
+static uint8_t floor_log2(uint16_t n);
+static uint16_t as_lower_case(uint8_t ch);
+static void generate_lfn(uint16_t *ucs, size_t ucs_max, const fs_utils_dir_entry_t *entry);
+static bool include_entry(fs_utils_dir_entry_flags_t entryFlags, fs_utils_dir_flags_t flags);
+
+
+// Copy a sequence of bytes; destination and source must _not_ overlap.
+static void copy_bytes(uint8_t *dst, const uint8_t *src, size_t len)
```
engdoreis: use memcpy from: https://github.com/lowRISC/mocha/blob/main/sw/device/lib/runtime/string.h

sw/device/lib/runtime/filesys_utils.c
```diff
@@ -0,0 +1,852 @@
...
+        len -= chunk_len;
+        bytesRead += chunk_len;
+    }
+    return bytesRead;
+}
+
+// public:
+// Unfortunately FAT stores the literal values for bytes/sector and sectors/cluster but only
+// powers of two are permitted.
+static inline uint8_t floor_log2(uint16_t n)
```
engdoreis: I think you can use a builtin_ctz here: https://github.com/lowRISC/mocha/blob/5a99b1d04bd5b12fb37e3803a12f2d0a3484f648/sw/device/lib/builtin.h#L14

sw/device/lib/runtime/filesys_utils.c
```diff
@@ -0,0 +1,852 @@
...
+
+    // Store pointers to SPI Host for SD comms and (optional) UART for logging
+    fs->spi = spi;
+    fs->uart = uart;
+
+    // Read the Master Boot Record (MBR) from block 0 at the very start of the medium.
+    uint8_t *dataBuffer = fs->buf.dataBuffer;
+    if (!read_blocks(fs->spi, 0, dataBuffer, 1u, fs->uart)) {
+        if (fs->uart) {
+            uart_puts(fs->uart, "Unable to read the MBR of the SD card\n");
```
engdoreis: I would use uprintf everywhere for consistency

sw/device/lib/runtime/filesys_utils.c
```diff
@@ -0,0 +1,852 @@
...
+fs_utils_dir_handle_t rootdir_open(fs_utils_state_t *fs)
+{
+    if (!fs->partValid) {
+        return FS_UTILS_INVALID_DIR_HANDLE;
+    }
+    return dir_open(fs, fs->rootCluster);
+}
+
+// Open a directory object that started in the given cluster.
+fs_utils_dir_handle_t dir_open(fs_utils_state_t *fs, uint32_t cluster)
```
engdoreis, line 373: This is a lot of code to maintain, for the purpose of testing the sdcard, handling directories is not important. So I suggest droping the directory handling and keeping the bare minimum flat file handling.
Specially because the real use case of the SDCard will be done in Uboot which already has a filesystem implementation.

sw/device/lib/runtime/sdcard.c
```diff
@@ -0,0 +1,505 @@
...
+    // Using only one cycle seemed to screw-up the next transaction, so we use eight.
+    nonblocking_cycles(spi, 8, false);
+}
+
+// Initialise the SD card ready for use.
+bool sdcard_init(spi_host_t spi, uart_t uart)
+{
+    // Every card tried seems to be more than capable of keeping up with 12.5Mbps.
+    const unsigned kSpiSpeed = 1u; // 50 MHz / (1 + 1) = 12.5 MHz
+    DEV_WRITE(spi + SPI_HOST_CONFIGOPTS_REG, 0xFFFF & kSpiSpeed);
```
engdoreis, line 29: Why not use the spi_host driver instead of directly accessing the mmios?

hw/top_chip/data/pins_genesys2.xdc
```diff
@@ -20,11 +21,13 @@ set_property -dict { PACKAGE_PIN N19   IOSTANDARD LVCMOS18 } [get_ports { gpio_i
 set_property -dict { PACKAGE_PIN P19   IOSTANDARD LVCMOS18 } [get_ports { gpio_i[5] }]; # SW5 (VADJ)
 set_property -dict { PACKAGE_PIN P26   IOSTANDARD LVCMOS33 } [get_ports { gpio_i[6] }]; # SW6 (VCC3V3)
 set_property -dict { PACKAGE_PIN P27   IOSTANDARD LVCMOS33 } [get_ports { gpio_i[7] }]; # SW7 (VCC3V3)
-
-# Bootstrap pin, should be pulled down during boot to enter bootstrap mode.
-set_property -dict { PACKAGE_PIN AB29  IOSTANDARD LVCMOS33 PULLTYPE PULLUP } [get_ports { gpio_i[8] }];
+#   Bootstrap pin, should be pulled down during boot to enter bootstrap mode.
+set_property -dict { PACKAGE_PIN AB29  IOSTANDARD LVCMOS33 PULLTYPE PULLUP } [get_ports { gpio_i[8] }]; # PROG_RXFN
```
marnovandermaas: What does "PROG_RXFN" mean?

hw/top_chip/dv/verilator/top_chip_verilator.sv
```diff
@@ -52,7 +57,7 @@ module top_chip_verilator (
     .clk_i,
     .rst_ni,
 
-    .gpio_i    (gpio_inputs),
+    .gpio_i    ({gpio_inputs[31:10], microsd_det, gpio_inputs[8:0]}),
```
marnovandermaas: This is a little bit ugly because the gpio_input now will just ignore any writes to bit index 9. I'm not sure how to solve it more cleanly though.

hw/top_chip/dv/verilator/top_chip_verilator.sv
```diff
@@ -30,20 +30,25 @@ module top_chip_verilator (
...
-  // SPI signals
+  // SPI device signals
   logic       spi_device_sck;
   logic       spi_device_csb;
   logic [3:0] qspi_device_sdo;
   logic [3:0] qspi_device_sdo_en;
   logic       spi_device_sdi;
 
+  // SPI host signals
+  logic       spi_host_sck_output,    spi_host_csb_output;
```
marnovandermaas: I'm not sure I think the "output" tag is more clear here. If you don't like just calling the signals spi_host_* then maybe we can label it h2d for host to device or something similar to indicate direction.

doc/ref/dev_guide.md
```diff
@@ -180,6 +180,8 @@ Then load the bitstream onto Genesys 2:
 openFPGALoader -b genesys2 build/lowrisc_mocha_chip_mocha_genesys2_0/synth-vivado/lowrisc_mocha_chip_mocha_genesys2_0.bit
 ```
 
+For the SD card tests, you will also need to insert a FAT32-formatted microSD card containing the "lorem.ips" file (see sw/device/tests/spi_host/lorem_text.h).
```
marnovandermaas: Nit: remove "will"

marnovandermaas (commented):
Thanks for working on this! I've left some initial comments.

hw/top_chip/data/pins_genesys2.xdc
elliotb-lowrisc: It keeps a record of how to connect up the other SD card pins, should they be needed in future.

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Are we not going to be adding support for asserts in the future?

sw/device/lib/runtime/sdcard.c
elliotb-lowrisc: The spi_host HAL does not presently support many of the features of the spi_host block that this code needs to make the SD card communications work. Perhaps the sdcard code could be re-written if the HAL is sufficiently improved, but until then I thought it better to just stick to using the spi_host block directly rather than only partially using the HAL.

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: I'm confused. This code is used in the SD Card test I added. Are you suggesting performing a lesser test?

In any case, I'd have thought it useful to keep this code in case we want to to use bare-metal SD card access in future, such as for a demo.

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Why use a formatting printer just to output a string?

sw/device/lib/hal/uart.c
elliotb-lowrisc: Wouldn't that create a circular reference?

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Thanks, I'll give this a go next time I have access to an FPGA

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: I think we want to find the leading set bit, not the trailing set bit

hw/top_chip/data/pins_genesys2.xdc
elliotb-lowrisc: It's the name in the Genesys2 schematic for the net connected to this pin. Unlike Sonata, we're not using the schematic names for the FPGA-top signal names (see https://github.com/lowRISC/mocha/issues/411), so I'm adding them as a comment instead.

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Indeed, I'm open to better ideas

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Ah right, I think I was following something like what I did for GPIO, but this is not bidirectional so things can be simpler

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Hmm, not sure then that'll be actually, so I'll try it in CI

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Better now?

elliotb-lowrisc:
Use `memcpy`, shorten some Verilator signal names, doc fix

sw/device/lib/runtime/filesys_utils.c
ziuziakowska: I believe floor log2 (sometimes called ilog2) would be `<bitwidth of x> - clz(x) - 1`, with an exception for `x = 0`. Those would be useful to have in `builtin.h`.

sw/device/lib/hal/uart.c
elliotb-lowrisc: I seem to have more capacity for problem solving today. I can move this function to a file of its own to get around the dependency issue, and buffer bytes into four-byte bundles to work around the lack of width/padding control in `uprintf` at present.

The result looks like this:
```
cd16cde4 54feeb19 20736968 6e207369 : .......This is n
6120746f 6f6f6220 6c626174 69642065 : ot a bootable di
202e6b73 656c5020 20657361 65736e69 : sk.  Please inse
```

Which seems good enough for now, if perhaps not as readable as the original intent:
```
E4 CD 16 CD 19 EB FE 54 68 69 73 20 69 73 20 6E : .......This is n
6F 74 20 61 20 62 6F 6F 74 61 62 6C 65 20 64 69 : ot a bootable di
73 6B 2E 20 20 50 6C 65 61 73 65 20 69 6E 73 65 : sk.  Please inse
```

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: If you'd both prefer it, I suppose I can switch to uprintf everywhere and put all non-error messages behind logging gating to avoid making what will be the slowest Verilator test even slower (roughly 10% slower from the switch before I added the extra gating)

elliotb-lowrisc:
Rework uart printing to use `uprintf` everywhere as requested

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Something like this?
```c
#define ilog2_16b(x) (x == 0 ? 0 : (16 - clz(x) - 1))
```

elliotb-lowrisc: Nope, that didn't work. I think I'm missing how to get the bitwidth of x as far as `clz` is concerned

elliotb-lowrisc:
Rebase

elliotb-lowrisc:
Lint fixes (I forgot to run the formatter after the last round of changes)

hw/top_chip/data/pins_genesys2.xdc
engdoreis: I would prefer not having commented lines as it polutes the file, the history record can always be recovered with git.

sw/device/lib/runtime/filesys_utils.c
engdoreis: For a simple string, the uprintf should perform almos the same, since there won't be args parse and format.
But I don't have strong opinion.

sw/device/lib/runtime/sdcard.c
engdoreis: I would than advocate to improve the spi_host HAL, to avoid duplicate code when these features are needed there.

sw/device/lib/runtime/filesys_utils.c
engdoreis: I believe that using directories in the filesystem would not increase any test converage of the hardware. As a reference in Opentitan we don't need to mount a filesystem to test that spi_host works properly with the flashes.
I'm just concern that with more code in the repo, the harder to debug tests the future and therefore I advocate for the bare minimum.

But I'm happy to keep it if others desagree.

hw/top_chip/data/pins_genesys2.xdc
elliotb-lowrisc: In this case, the lines did not exist before so they could not be recovered from the Git history if they are not included in this commit

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: I've done the switch to uprintf now, so might as well stick with it

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: The test currently uses it in order to open the root directory so that it can read the "lorem.ips" test files. To remove the ability to open files would be rather limiting

sw/device/lib/runtime/sdcard.c
elliotb-lowrisc: How about if I moved sdcard.h/c from the "runtime" directory to the "hal" dir? If we are only going to be using the spi_host to talk to an SD card, then it feels like taking the time to add another layer of abstraction between the spi_host and the SD card code should be a low-priority right now

hw/top_chip/data/pins_genesys2.xdc
elliotb-lowrisc: That said, this info can be found from the Genesys 2 schematic and SD card specification again in future if it si needed, so I suppose I could just remove these lines on the assumption that we are unlikely to use them

elliotb-lowrisc:
Rebase and tweak documentation change to better match new dev guide section on additional hardware

hw/top_chip/dv/verilator/top_chip_verilator.sv
marnovandermaas: The only thing I can think of is that it is probably better to put the microSD detection in the most significant bit rather than in the middle. What do you think?

hw/top_chip/dv/verilator/top_chip_verilator.sv
marnovandermaas: Yes, looks good.

hw/top_chip/dv/verilator/top_chip_verilator.sv
ziuziakowska: If you do change the GPIO, make sure to update this line in the devicetree and regenerate it: https://github.com/lowRISC/mocha/blob/main/sw/device/devicetree/mocha.dts#L76

sw/device/lib/runtime/filesys_utils.c
ziuziakowska: https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html says that `__builtin_clz` takes an unsigned int, which on RISC-V is always 32-bits, so it should be `(32 - clz(x) - 1)`. It should be fine to put a `uint16_t` into that.

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Introducing a gap in the range of GPI lines in use might make things trickier on the FPGA-side, where at the moment the input signal is only wide enough for the number of GPI lines assigned to pins. We could change it so that the FPGA GPI input is 32-bits wide and only assign some a subset to pins in the XDC file, or we could adopt https://github.com/lowRISC/mocha/issues/411 and have the mapping of board signals to GPI lines in the RTL rather than constraints, where it would be easier to assign some constant values.

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Ah, that seems to work, but I'm not sure what to call it. My first inclination is `ilog_32b` due to it (I assume) not working on values wider than 32 bits, but perhaps `ilog_u` (for unsigned int) would be more appropriate. Do you know of any convention to follow in such situations?

ziuziakowska: Yes this won't work on 64 bit values, so a separate one for those could be added that has `64` and `clzll(x)` in those same spots. It would be great if you could add these as `static inline` functions in `builtin.h`.

C Convention uses no suffix for (unsigned)`int` variants of maths functions, `l` for `long` variants, and `ll` for `long long` variants. On RISC-V, `int` is always `uint32_t` and `long long` is always `uint64_t`, so `ilog2` and `ilog2ll` with those types should do.

The unsigned-ness is implied by this being a logarithm function as those are not defined for negative numbers, and should be made explicit in the function parameter types.

elliotb-lowrisc: Alright, I've hopefully now implemented what you have been guiding me towards

elliotb-lowrisc:
Rework log2 to use builtins

elliotb-lowrisc:
Rebase to resolve RTL conflicts

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: To expand upon what I said (in the hope of making some progress), dropping the filesystem code would still allow us to test whether the spi_host can initialise and read the CSD and CID registers of an SD card or SD card model, just not the commands used to read data stored in the main storage area. As the process for reading the aforementioned registers is not too dissimilar from reading the main storage the spi_host HW would still be exercised, just not with as large a set of read data. What would be lost is the ability to read files from an SD card as part of a bare-metal application/demo.

@marnovandermaas do you have a view on this topic that could help us make a decision?

sw/device/lib/runtime/sdcard.c
elliotb-lowrisc: @engdoreis What do you think of my proposal?

sw/device/lib/runtime/filesys_utils.c
engdoreis: I had a quick chat with @marnovandermaas about this, and we think that it makes sense to go ahead and merge the full FS teting for now.
My thinking is that once we integrate U-Boot, we can test the SD card using a U-Boot script and eventually drop the bare-metal FS. Since we don't currently have a dedicated bare-metal strategy for Mocha, any bare-metal needs down the road could probably be handled with an out-of-tree FS instead.
Does that sound reasonable to you, or do you see any issue with this approach?

sw/device/lib/runtime/sdcard.c
engdoreis: I'm aligned on keeping the abstraction simple and low-priority for now. However, the hal directory should strictly be reserved for SoC drivers.
Since the SD card code uses the SPI host to communicate, it belongs a layer above. I think it should live in a drivers/ directory instead but runtime is also ok. This keeps a clean separation of responsibilities across layers.

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Sure, I'm OK with that. My main aim is to get this merged. If parts are removed later, then so be it

sw/device/lib/runtime/sdcard.c
elliotb-lowrisc: Sure, I can move it to a new "drivers/" directory

elliotb-lowrisc: It seems I can't without removing `uprintf` due to the circular dependency it introduces, ah well

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: @engdoreis do you have any new thoughts on this now that the future of this file is clearer?

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: @marnovandermaas your thoughts would be appreciated

elliotb-lowrisc:
Removed commented-out XDC lines (they can always be re-created from the schematic)

hw/top_chip/data/pins_genesys2.xdc
elliotb-lowrisc: Now removed

sw/device/lib/runtime/filesys_utils.c
engdoreis: I think that any comment code should be removed. A `TODO` command can be used to remind that asserts should be added.

elliotb-lowrisc: Alright, I guess if this file is likely to be deleted, then it is unlikely that there will be an opportunity to enable these asserts

elliotb-lowrisc:
Removed commented-out assertion lines from filesys code, and removed/TODO-ed in sdcard code

sw/device/lib/runtime/filesys_utils.c
elliotb-lowrisc: Alright, I've removed the assert stuff from this file, and removed or added TODOs to assert stuff in sdcard.c. Does that work for you?

hw/top_chip/dv/verilator/top_chip_verilator.sv
marnovandermaas: I think I would go for a gap where the most significant bits are ones you cannot affect by the DPI and any future general purpose inputs that can be affected by the DPI can be added contiguously at the bottom. I'm happy for you to also implement the #411 at the same time if you think that's not too much work.

elliotb-lowrisc: Alright, I'll move the input to a higher index. I might do the rest of the FPGA pin remapping an a subsequent PR.

@ziuziakowska Is there documentation you could point me to for how to re-generate the device tree?

ziuziakowska: The command is here: https://github.com/lowRISC/mocha/blob/main/util/artefacts.py#L47-L58

engdoreis (approved):

elliotb-lowrisc:
Changed GPIO input used for SD card detect, rebased to avoid merge conflict

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Done thanks

hw/top_chip/dv/verilator/top_chip_verilator.sv
```diff
@@ -57,7 +62,7 @@ module top_chip_verilator (
     .clk_i,
     .rst_ni,
 
-    .gpio_i    (gpio_inputs),
+    .gpio_i    ({microsd_det, gpio_inputs[30:0]}),
```
marnovandermaas, line 65: Not necessary for this PR, but maybe good to open an issue. It would be good to have an assertion on gpio_inputs[31] that if this value ever changes that it prints a warning during simulation.

hw/top_chip/dv/verilator/top_chip_verilator.sv
```diff
@@ -82,17 +87,15 @@ module top_chip_verilator (
...
-    // sd_i[1]=CIPO.
-    .spi_host_sd_i     ({2'b0, spi_host_sd_en[0] ? spi_host_sd[0] : 1'b0, 1'b0}),
+    .spi_host_sck_o    (spi_host_sck),
+    .spi_host_sck_en_o (spi_host_sck_en),
+    .spi_host_csb_o    (spi_host_csb),
+    .spi_host_csb_en_o (spi_host_csb_en),
+    .spi_host_sd_o     (spi_host_sd_output),
+    .spi_host_sd_en_o  (spi_host_sd_en_output),
+    // Legacy SPI present in SD cards does not allow bi-directional wires.
+    // Work in standard mode where sd_o[0]=COPI and sd_i[1]=CIPO.
```
marnovandermaas: "Work" -> "Only works"

hw/top_chip/rtl/chip_mocha_genesys2.sv
```diff
@@ -164,6 +178,10 @@ module chip_mocha_genesys2 #(
 
   assign fpga_rst_n_sync_cfg = fpga_rst_n_shreg[0];
 
+  // SD Card reset (active-low) unused for now.
+  // Connect up to GPIO output if have SD card reliability issues.
```
marnovandermaas: "if have" -> "if we have"

.gitignore
```diff
@@ -27,3 +27,6 @@ __pycache__
 scratch/
 # clangd
 .cache
+
+# SD card image file
+sd.img
```
marnovandermaas, line 32: Are there any instructions on how to generate this image?

marnovandermaas: Also how does Verilator know the content of this image file?

marnovandermaas (commented):
Some minor comments on my end.

.gitignore
elliotb-lowrisc: @marnovandermaas There are some instructions in sonata-system, though they only really cover how to download an existing image and add a file to it. The SD card DPI model used in the Verilator testbench just opens the image as a normal binary file (see spi_microsd.hh/cc).

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: Issue created: https://github.com/lowRISC/mocha/issues/585

elliotb-lowrisc:
Improve highlighted code comments and add link to sonata-system SD card image instructions

hw/top_chip/dv/verilator/top_chip_verilator.sv
elliotb-lowrisc: @marnovandermaas are you happy with my improved comment?

hw/top_chip/rtl/chip_mocha_genesys2.sv
elliotb-lowrisc: @marnovandermaas are you happy with my improved comment?

.gitignore
elliotb-lowrisc: @marnovandermaas I've added a link to the sonata-system instructions to the dev guide

sw/device/lib/builtin.h
```diff
@@ -4,14 +4,30 @@
...
+#define ctzll(x) (__builtin_ctzll((x)))
+/* count leading zeroes */
+#define clz(x)   (__builtin_clz((x)))
+#define clzll(x) (__builtin_clzll((x)))
+
+/* floor log2 */
+static inline uint32_t ilog2(uint32_t x)
+{
+    return (x) == 0u ? 0u : (32u - clz(x) - 1u);
+}
```
ziuziakowska, line 26: Nit: Functions should be separated by a newline.

sw/device/lib/builtin.h
```diff
@@ -4,14 +4,30 @@
...
+#define ctz(x)   (__builtin_ctz((x)))
+#define ctzll(x) (__builtin_ctzll((x)))
+/* count leading zeroes */
+#define clz(x)   (__builtin_clz((x)))
+#define clzll(x) (__builtin_clzll((x)))
+
+/* floor log2 */
+static inline uint32_t ilog2(uint32_t x)
+{
+    return (x) == 0u ? 0u : (32u - clz(x) - 1u);
```
ziuziakowska: Nit:
```suggestion
    return (x == 0u) ? 0u : (32u - clz(x) - 1u));

**Review by ziuziakowska — APPROVED** [review]
LGTM. Just some nits, and I think the commit that makes the devicetree changes should be squashed with the first.

File: sw/device/lib/builtin.h (hunk 19 above)
Comment (elliotb-lowrisc) [reply sw/device/lib/builtin.h]: There doesn't seem to be a balanced number of parentheses in the proposed change. Did you mean?:
```
return (x == 0u) ? 0u : (32u - clz(x) - 1u);
```

Comment (ziuziakowska) [reply sw/device/lib/builtin.h]: You're right there's one too many, yeah

**Comment by elliotb-lowrisc** [comment]
Resolved builtin.h nits

**Comment by elliotb-lowrisc** [comment]
Fix reset polarity, remove some changes leftover from earlier iterations

**Comment by elliotb-lowrisc** [comment]
Please forgive yet another push to fix a minor issue, this time a SW include

File: hw/top_chip/dv/verilator/top_chip_verilator.sv (hunk 14 above)
Comment (marnovandermaas) [reply hw/top_chip/dv/verilator/top_chip_verilator.sv:65]: Thank you that is great!

File: .gitignore (hunk 17 above)
Comment (marnovandermaas) [reply .gitignore:32]: Thanks Elliot this makes sense. And it also makes sense that Verilator does not need this image since it only does some register reads in this test.

File: hw/top_chip/dv/verilator/top_chip_verilator.sv [hunk 20]
```diff
@@ -108,8 +111,7 @@ module top_chip_verilator (
   );
 
   // No support for dual or quad SPI in loopback mode right now.
-  logic unused_spi_host = (|spi_host_sd[3:2]) | spi_host_sd[0] |
-                          (|spi_host_sd_en[3:2]) | spi_host_sd_en[0];
+  logic unused_spi_host = |{spi_host_sd_output[3:1], spi_host_sd_en_output[3:1]};
```
marnovandermaas, line 114: Good catch, this was a mistake from before.

.gitignore
elliotb-lowrisc: For clarity, the presence or absence of an sd.img file is signalled by the DPI model using the CID register:
https://github.com/lowRISC/mocha/blob/1c9e9dadd8aa7e0548ac8740f56bbab87484ac55/hw/vendor/sonata_system/dv/dpi/spidevicedpi/spi_microsd.cc#L56-L63

Which is then used to determine whether to run the file system part of the test:
https://github.com/lowRISC/mocha/pull/446/changes#diff-68e3d4da915a4913777990063980f6f73375598ed41b508be7aaa170b03f9472R175-R180

sw/device/tests/spi_host/lorem_text.h
```diff
@@ -0,0 +1,85 @@
+
```
marnovandermaas: Nit: we don't need to start with an empty line right?

marnovandermaas (approved):
One more minor comment but otherwise good to go. I ran it locally with logging and seems to do what I expect.

sw/device/tests/spi_host/lorem_text.h
elliotb-lowrisc: No, that looks like an oversight on my part. I'll fix it

elliotb-lowrisc:
Delete unnecessary blank line

sw/device/tests/spi_host/lorem_text.h
elliotb-lowrisc: Done, could I get a quick re-approval from you?

marnovandermaas (approved):
Reapprove after just a single line fix.

PR #447, opened 2026-04-18
Move the bootrom to the ROM

engdoreis:
This PR:
 - Move the bootrom to the ROM
 - Move all tests to run on DRAM rather than SRAM.
 - Enable all verilator tests to use the bootrom.


hw/vendor/patches/lowrisc_ip/gpiodpi/001_enable_weak_pins_by_default.patch
marnovandermaas, line 1: Nit: we usually use 4 digits here so `0001`

sw/cmake/tests.cmake
```diff
@@ -13,21 +13,24 @@ set(
 
 # for a given executable, create a raw binary (.bin), verilog memory (.vmem),
 # and disassembly output for debugging (.dump).
-function(mocha_add_executable_artefacts NAME)
+function(mocha_add_executable_artefacts)
+  set(one_value_args NAME)
```
marnovandermaas: Did you intend to only indent this by two spaces? Is that a convention for defining arguments?

hw/top_chip/rtl/chip_mocha_genesys2.sv
```diff
@@ -4,7 +4,7 @@
 
 module chip_mocha_genesys2 #(
   parameter BootRomInitFile = "",
-  parameter RomInitFile     = ""
+  parameter SramInitFile    = ""
```
marnovandermaas: I prefer if we call these parameter SramInitFile and RomInitFile. Then the Boot ROM image can be put into the RomInitFile.

hw/top_chip/rtl/chip_mocha_genesys2.sv
```diff
@@ -148,8 +148,8 @@ module chip_mocha_genesys2 #(
 
   // CHERI Mocha top
   top_chip_system #(
-    .SramInitFile(BootRomInitFile),
-    .RomInitFile (RomInitFile)
+    .SramInitFile(SramInitFile),
+    .RomInitFile(BootRomInitFile)
```
marnovandermaas: Nit: alignment

sw/cmake/tests.cmake
```diff
@@ -35,12 +35,13 @@ endfunction()
 
 # for a given executable, add a test that runs the executable
 # in the Verilator simulation.
-function(mocha_add_verilator_test NAME)
+function(mocha_add_verilator_test NAME BOOTROM)
+  # Warning: If the BootROM and the test are compiled to the same memory address, the test will 
+  # effectively replace the BootROM. Because the BootROM is listed first in the arguments, it is
+  # overwritten by the subsequent test image; the simulation will behave as if no BootROM is present
```
marnovandermaas: Did you mean for this to be indented by only 2 spaces?

sw/cmake/tests.cmake
```diff
@@ -35,21 +35,43 @@ endfunction()
...
-        COMMAND ${PROJECT_SOURCE_DIR}/../util/verilator_runner.sh -E $<TARGET_FILE:${BOOTROM}> -E ${NAME}
+        COMMAND ${PROJECT_SOURCE_DIR}/../util/verilator_runner.sh -E $<TARGET_FILE:${arg_ROM}> -E ${arg_NAME}
     )
+    set_tests_properties(${TEST} PROPERTIES TIMEOUT ${arg_TIMEOUT})
+
 endfunction()
 
-function(mocha_add_fpga_test NAME)
+function(mocha_add_fpga_test)
+    set(one_value_args NAME ROM TIMEOUT)
```
marnovandermaas: Is ROM used in this function?

hw/top_chip/dv/top_chip_sim_cfg.hjson
marnovandermaas, line 1: We need to make sure to run and check this before merging.

sw/cmake/tests.cmake
```diff
@@ -96,30 +92,28 @@ function(mocha_add_test)
     foreach(ARCH_NAME FLAGS_VAR IN ZIP_LISTS ARCHS ARCHS_FLAGS)
       set(FLAGS ${${FLAGS_VAR}})
 
-      foreach(CONFIG OFFSET FPGA SIM IN ZIP_LISTS BOOT_CFG BOOT_CFG_OFFSET BOOT_CFG_FPGA BOOT_CFG_VERILATOR)
```
marnovandermaas, line 73: Nit: You're removing the foreach but not adjusting the indentation.

marnovandermaas (commented):
This is a very cool PR. Just a few nits from my end and we need to make sure to test the DV run to make sure nothing breaks.

sw/device/tests/CMakeLists.txt
ziuziakowska, line 1: Something to think about: Maybe we could flip the condition to be `NO_FPGA`/`SKIP_FPGA` as that could stand out more, and it makes more sense with comments/TODOs explaining why that is there instead of why `FPGA` isn't there.

hw/top_chip/dv/verilator/top_chip_verilator.sv
```diff
@@ -107,8 +107,8 @@ module top_chip_verilator (
     .gpio_p2d      (gpio_inputs),
     .gpio_d2p      (gpio_outputs),
     .gpio_en_d2p   (gpio_en_outputs),
-    .gpio_pull_en  ('0), // no pull-ups in simple SoC
-    .gpio_pull_sel ('0)  // no pull-ups in simple SoC
+    .gpio_pull_en  (32'hFFFF_FFFF), // pull-ups for all GPIOs
+    .gpio_pull_sel (32'hFFFF_FFFF)  // pull-ups for all GPIOs
```
ziuziakowska, line 111: I'm not the most familiar with this, what is the reasoning for this change? I had some issues with the boot ROM in Verilator going into SPI bootstrap and had to drive the GPIO to bypass it, so if this is to fix that, there should be a commit message or comment in the boot ROM with the rationale.

sw/device/lib/boot/mocha_dram.ld
```diff
@@ -0,0 +1,19 @@
+/* Copyright lowRISC contributors (COSMIC project). */
+/* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+OUTPUT_ARCH(riscv)
+
+INCLUDE memory.ld
+
+/* TODO: Remove this when we implement a ROM. */
+/* This should be provided by the build system, otherwise the default is 0x0000*/
```
ziuziakowska: nit: spacing (also for `mocha_sram.ld`)
```suggestion
/* This should be provided by the build system, otherwise the default is 0x0000 */
```

sw/device/lib/boot/mocha.ld
ziuziakowska, line 1: Maybe we could have a `lib/boot/linkerscript` or `lib/boot/ld` subdirectory to contain these, as there are quite a few now.

sw/device/bootrom/bootrom.c
```diff
@@ -77,6 +79,14 @@ void boot(uintptr_t addr)
     reset();
 }
 
+void clear_slots()
+{
+    for (size_t i = 0; i < ARRAY_LEN(boot_slots); i++) {
+        uintptr_t slot = boot_slots[i];
+        DEV_WRITE(slot, 0x00);
```
ziuziakowska: I would prefer
```suggestion
        DEV_WRITE(boot_slots[i], 0);
```

sw/device/bootrom/rom.ld
```diff
@@ -35,24 +36,24 @@ SECTIONS {
...
         *(.sdata.*)
         *(.data)
         *(.data.*)
         . = ALIGN(8);
-    } > ram
+
+        _data_end = .;
+        _data_init_end = LOADADDR(.data) + SIZEOF(.data);
+
+    } > ram AT> rom
```
AlexJones0: Nit:
```suggestion
    } > ram AT > rom
```

sw/device/bootrom/bootrom.c
```diff
@@ -53,6 +54,7 @@ int main(void)
     timer_enable_write(boot_ctx.timer, true);
     if (bootstrap_requested(&boot_ctx)) {
         uprintf(boot_ctx.console, "Entering SPI bootstrap\n");
+        clear_slots(); // if a binary is loaded into slot 1, we don't want slot 0 to boot.
```
ziuziakowska: Maybe this comment should say it is just clearing memory from a previous boot.

sw/device/bootrom/CMakeLists.txt
```diff
@@ -8,9 +8,30 @@ target_link_libraries(${NAME} hal_vanilla runtime_vanilla startup_vanilla)
...
+    COMMAND ${CMAKE_OBJDUMP} ${OBJDUMP_FLAGS} "$<TARGET_FILE:${NAME}>"
+            > "$<TARGET_FILE:${NAME}>.dump"
+
+    COMMAND ${CMAKE_OBJCOPY} -O binary
+      -R .bss -R .stack  # already excluded (NOBITS), but for clarity
+      --only-section=.init_vectors
+      --only-section=.text
+      --only-section=.rodata
+      --only-section=__cap_relocs
+      --only-section=.data   # LMA is in ROM, objcopy uses LMA when --only-section forces i
```
AlexJones0:
```suggestion
      --only-section=.data   # LMA is in ROM, objcopy uses LMA when --only-section forces it
```

sw/device/bootrom/rom.ld
```diff
@@ -0,0 +1,88 @@
+/* Copyright lowRISC contributors (COSMIC project). */
+/* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+OUTPUT_ARCH(riscv)
+
+INCLUDE memory.ld
+
+MEMORY {
+    rom   (RWX) : ORIGIN = ORIGIN(ROM) , LENGTH = (LENGTH(ROM))
```
ziuziakowska: Should ROM be marked as `RX`?

hw/top_chip/rtl/chip_mocha_genesys2.sv
```diff
@@ -3,8 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 module chip_mocha_genesys2 #(
-  parameter BootRomInitFile = "",
-  parameter RomInitFile     = ""
+  parameter RomInitFile = "",
+  parameter SramInitFile    = ""
```
AlexJones0: Nit: formatting/spacing here is inconsistent.

sw/device/bootrom/CMakeLists.txt
```diff
@@ -8,9 +8,30 @@ target_link_libraries(${NAME} hal_vanilla runtime_vanilla startup_vanilla)
...
+      --only-section=.got
+      "$<TARGET_FILE:${NAME}>"
+      "$<TARGET_FILE:${NAME}>.bin"
```
ziuziakowska, line 22: The objcopy manpage warns against using these as it may make the file unusable, so this may be a footgun if we do not remember to update this here. Is there a reason this is specifically needed over just using `objcopy -O binary`?

hw/vendor/cva6_cheri/core/include/cv64a6_imafdchzcheri_sv39_config_pkg.sv
```diff
@@ -140,9 +140,9 @@ package cva6_config_pkg;
       NrNonIdempotentRules: unsigned'(2),
       NonIdempotentAddrBase: 1024'({64'b0, 64'b0}),
       NonIdempotentLength: 1024'({64'b0, 64'b0}),
-      NrExecuteRegionRules: unsigned'(3),
-      ExecuteRegionAddrBase: 1024'({64'h8000_0000, 64'h1000_0000, 64'h2000_0000}),
-      ExecuteRegionLength: 1024'({64'h4000_0000, 64'h2_0000, 64'h1000}),
+      NrExecuteRegionRules: unsigned'(4),
+      ExecuteRegionAddrBase: 1024'({64'h8000_0000, 64'h1000_0000, 64'h2000_0000, 64'h0008_0000}),
+      ExecuteRegionLength: 1024'({64'h4000_0000, 64'h2_0000, 64'h1000, 64'h8000}),
```
AlexJones0: Question (not familiar with this bit of code). I guess the duplication in the vendored CVA6 RTL here (writing `64'h0008_0000` and `64'h8000` instead of `top_pkg::RomCtrlMemBase` and `top_pkg::RomCtrlMemLength`) are because the patches live in our fork which cannot know about the mocha `top_pkg.sv`?

I'm not sure I have a better solution, but this seems like it would be quite fragile/difficult to move things around in the address space if needed. Is there any way we could at least have some assertions to catch this kind of thing?

util/verilator_runner.sh
ziuziakowska, line 1: Doesn't need to be done in this PR, but now we can remove this runner and have Verilator tests run the binary directly.

sw/device/tests/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 set(LIBS hal runtime startup test_framework)
 
-mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME test_framework_test TIMEOUT 20 SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} FPGA)
```
ziuziakowska: I think the `TIMEOUT` could go at the end of the list with the test environments.

ziuziakowska (commented):
These are really great changes, I just have some nits and comments.

sw/cmake/tests.cmake
```diff
@@ -13,94 +13,112 @@ set(
...
-macro(mocha_add_verilator_test NAME)
+function(mocha_add_verilator_test)
+    # Warning: If the BootROM and the test are compiled to the same memory address, the test will 
+    # effectively replace the BootROM. Because the BootROM is listed first in the arguments, it is
+    # overwritten by the subsequent test image; the simulation will behave as if no BootROM is present
+
+    set(one_value_args NAME ROM TIMEOUT)
+    cmake_parse_arguments(arg "" "${one_value_args}" "" ${ARGN})
+
+    if(NOT arg_TIMEOUT )
```
AlexJones0:
```suggestion
    if(NOT arg_TIMEOUT)
```

sw/device/tests/CMakeLists.txt
```diff
@@ -23,5 +23,5 @@ mocha_add_test(NAME tag_controller_tag_test SOURCES tag_controller/tag_test.c LI
 mocha_add_test(NAME timer_smoketest SOURCES timer/smoketest.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME timer_interrupt_test SOURCES timer/interrupt.c LIBRARIES ${LIBS} FPGA)
 mocha_add_test(NAME uart_smoketest SOURCES uart/smoketest.c LIBRARIES ${LIBS} FPGA)
+mocha_add_test(NAME isa_atomics SOURCES isa/atomics.c LIBRARIES ${LIBS} FPGA SKIP_SRAM)
```
AlexJones0: Does this `SKIP_SRAM` do anything?

I can't seem to find any other references. Perhaps I'm missing something but it seems like it might be stale?

AlexJones0 (approved):
LGTM, with a few small nits & questions.

sw/device/lib/boot/mocha_dram.ld
engdoreis: The comment was removed

util/verilator_runner.sh
engdoreis: I like the idea, but rather than fixing the path to verilator, we should use cmake to find it. So I'll leave it to a next PR.

hw/top_chip/rtl/chip_mocha_genesys2.sv
engdoreis: good catch, left over from a search and replace.

hw/vendor/cva6_cheri/core/include/cv64a6_imafdchzcheri_sv39_config_pkg.sv
engdoreis: Yes, this configuration lives in a different repository, that's why we repeat these constants.
I don't know a better solution since I'm not very falmilar with sv. Maybe @marnovandermaas  can give you a better answer.

sw/device/tests/CMakeLists.txt
engdoreis: Left over, thanks for catching this.

sw/device/bootrom/CMakeLists.txt
engdoreis: This was to workarround the large binary size caused by the gap in the  address space between data and bss.
I fixed it by marking bss and stack as NOLOAD in the linker script.

martin-velay:
@thommythomaso, can you take a look at the RTL change of this commit: "[hw] Fix error sram error in dv"

sw/device/tests/CMakeLists.txt
marnovandermaas: I think this is a good idea.

hw/vendor/cva6_cheri/core/include/cv64a6_imafdchzcheri_sv39_config_pkg.sv
raylau1: You can override CVA6 parameters in `top_chip_system` by modifying the function `build_cva6_config`. Perhaps you can add something like:
```systemverilog
    cfg.NrExecuteRegionRules  = unsigned'(3);
    cfg.ExecuteRegionAddrBase = 1024'({top_pkg::DRAMBase, top_pkg::SRAMBase, top_pkg::RomCtrlMemBase});
    cfg.ExecuteRegionLength   = 1024'({top_pkg::DRAMLength, top_pkg::SRAMLength, top_pkg::RomCtrlMemLength});
    cfg.NrCachedRegionRules   = unsigned'(1);
    cfg.CachedRegionAddrBase  = 1024'({top_pkg::DRAMBase});
    cfg.CachedRegionLength    = 1024'({top_pkg::DRAMLength});
```

hw/top_chip/dv/verilator/top_chip_verilator.sv
engdoreis: I've improved the commit message.

hw/vendor/cva6_cheri/core/include/cv64a6_imafdchzcheri_sv39_config_pkg.sv
marnovandermaas, line 1: This patch can now be removed due to the DRAM fix PR.

marnovandermaas (commented):
This all looks good to me. Let's make sure that the top-level DV passes before merging but after that I'm happy for this to be merged.

sw/device/tests/CMakeLists.txt
engdoreis: This PR is tool large already, so I opened an issue: https://github.com/lowRISC/mocha/issues/477

hw/vendor/cva6_cheri/core/include/cv64a6_imafdchzcheri_sv39_config_pkg.sv
engdoreis: Marno has done what Ray suggested in main, so dropped this patch.

AlexJones0 (approved):
LGTM, thanks for the work on this.

I'm assuming the SRAM HW change is correct in that we cannot have `sram_we` when we are trying to read data, but I don't know enough about mocha's SRAM to validate that myself.

util/build_sw_collateral_for_sim.py
```diff
@@ -18,7 +18,7 @@ def generate(args) -> None:
 
     build_cmd = ["cmake", "--build", build_dir, "-v", "--target"]
     install_cmd = ["cmake", "--install", build_dir, "--prefix", out_dir, "--component"]
-    for img in args.sw_images:
+    for img in args.sw_images.split():
```
AlexJones0, line 21: Nit: if it can be done quickly, we should look into what part of DVSim / the HJSON flows is requiring us to do this and fix it there instead. This isn't a problem at all since we're running on Linux but is a bit of a code smell generally if it can be avoided.

AlexJones0 (approved):

PR #454, opened 2026-04-22
[rdl] ROM and DRAM sizes fixed

marnovandermaas:
The size of ROM and DRAM were wrong in multiple places. This fixes them in RDL and the Verilator simulation.


hw/top_chip/dv/verilator/top_chip_verilator.cc
```diff
@@ -93,11 +93,11 @@ bool MochaSim::Finish() {
 int main(int argc, char **argv) {
   MochaSim mocha_sim(
       "TOP.top_chip_verilator.u_top_chip_system.u_axi_sram.u_ram",
-      16 * 1024, // 16k 64-bit words = 128 KiB
+      16 * 1024, // 16K 64-bit words = 128 KiB
```
engdoreis, line 96: Todo in a follow PR, we could generate a header from SystemRDL with this values to be imported here.

engdoreis (approved):

hw/top_chip/dv/verilator/top_chip_verilator.cc
marnovandermaas: Tracked here: https://github.com/lowRISC/mocha/issues/459

PR #457, opened 2026-04-22
Tools specification

### LINKED ISSUE #439: Define tool versions
Define all the tool versions that we use in Mocha. For example, Verilator, Xcelium and Vivado.

marnovandermaas:
What tools do we need for Mocha and validate this against our schema.

Closes: https://github.com/lowRISC/mocha/issues/439


marnovandermaas:
@ziuziakowska mentioned that we should probably use a Python library to validate this: https://python-jsonschema.readthedocs.io/en/stable/validate/
This avoids adding Node as a dependency.

flake.nix
```diff
@@ -80,12 +80,14 @@
         gtkwave
         openfpgaloader
         openocd
+        nodejs
```
engdoreis: I'm not very keen on adding nodejs as a dependency only for the json schema validation. I suggesting using either python or compiled tools, a few options are:
 - jsonschema-cli: a Rust binary and can be installed with nix.
 - check-jsonschema: a Python script that can be installed with uv.
 - Our own script based on [opentitan json schema](https://github.com/lowRISC/opentitan/blob/b76d9ada848d1d78db878892d051b1c44f7ffcfe/util/validate_testplans.py) linter.

engdoreis (commented):
This is a good feature but I left a comment about the dependencies.

flake.nix
marnovandermaas: I've gone for the Python checking although I didn't know about the check-jsonschema.

marnovandermaas: The check-jsonschema script doesn't allow downloading a JSON file from the internet. Not sure if you would prefer a wget and then a check-jsonschema call or the Python script that I have now prepared in the utilities folder.

engdoreis: Using wget would requirer a wrapper script? Then It would be better to use check-jsonschema as a python lib and use request to donwload the JSON.

marnovandermaas: Are you happy with my current script that uses `jsonschema` directly or should I move to `check-jsonschema`?

package.json
ziuziakowska, line 1: Since you're not using nodejs anymore this file can be removed.

tool_data.json
```diff
@@ -0,0 +1,34 @@
...
+    {
+        "name": "Jasper",
+        "vendor": "Cadence",
+        "version": "2025.09p002",
+        "comment": "Formal verification."
+    },
+    {
+        "name": "Verilator",
+        "vendor": "Veripool",
+        "version": "v5.0.40",
```
ziuziakowska:
```suggestion
        "version": "v5.040",

File: package.json
Comment (marnovandermaas) [reply package.json:1]: Thanks, sorry for not removing this earlier.

File: tool_data.json (hunk 2 above)
Comment (marnovandermaas) [reply tool_data.json]: Good catch, this should now be fixed.

File: tool_data.json [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+        "name": "Vivado",
+        "vendor": "Xilinx",
+        "version": "v2021.1",
+        "comment": "Used to build bitstreams for Genesys 2."
+    },
+    {
+        "name": "Xcelium",
+        "vendor": "Cadence",
+        "version": "24.03-s007",
+        "comment": "Provides xrun for simulations and is used to run our full DV environment."
```
engdoreis:
```suggestion
        "comment": "Provides xrun for simulations and is used to run our full UVM environment."
```

engdoreis (approved):

tool_data.json
marnovandermaas: Thanks, that's now fixed.

PR #403, opened 2026-04-07
Build OpenSBI

ziuziakowska:
This PR adds a build step for OpenSBI - an M-mode firmware and "bootloader" - as well as associated changes to the boot ROM for it. The https://github.com/lowRISC/opensbi/tree/mocha-devel fork is used which adds drivers for enough Opentitan devices to get it printing and booting. The firmware is built with a test payload in the Mocha tree that performs an SBI call to print the test success message for FPGA platforms, and to report success to the DV status register.

This PR also adds a basic devicetree hardware description of Mocha, which is baked-in to the ROM. The bindings used should be kept in sync with our OpenSBI fork. 

The boot ROM is modified to pass this pointer in the `a1` register, and the hart id (0) in the `a0` register, as expected by OpenSBI.

The boot ROM is also modified to jump to the DRAM base by default if no boot magic number is found, as modifying the linkerscript of OpenSBI to include the magic number and boot address caused a lot of issues, and it means next stage software does not need intrusive modification to run.

PR #461, opened 2026-04-22
Nightly run for OpenSBI tests

### LINKED ISSUE #434: [CI] Create nightly run using github actions to run OpenSBI in verilator
Use one of the nix runners to run every day at 00:00 am.

ziuziakowska:
This adds OpenSBI tests for Verilator and FPGA, using the test payload in https://github.com/lowRISC/mocha/pull/403, and adds a Nightly pipeline to run the slow Verilator test - which takes upwards of 30-40m on regular hardware.

The FPGA test is fast enough to run as part of regular testing.

Closes https://github.com/lowRISC/mocha/issues/434.


machshev (approved):
Thanks @ziuziakowska!

sw/device/tests/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 set(LIBS hal runtime startup test_framework)
 
-mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS} TIMEOUT 20)
+mocha_add_test(NAME test_framework_test SOURCES test_framework/smoketest.c LIBRARIES ${LIBS})
```
engdoreis: Please keep the argument used at least once to make sure it won't break in the future.

sw/cmake/opensbi.cmake
```diff
@@ -77,4 +77,29 @@ function(mocha_opensbi_with_payload PAYLOAD_TARGET)
...
+# TODO: Integrate this with the build system better.
+function(mocha_opensbi_with_payload_test PAYLOAD_TARGET)
+  # name of the target OpenSBI build.
+  set(NAME opensbi_with_${PAYLOAD_TARGET})
+
+  mocha_opensbi_with_payload(${PAYLOAD_TARGET})
+
+  add_test(
+      NAME ${NAME}_sim_verilator
+      COMMAND ${PROJECT_SOURCE_DIR}/../util/verilator_runner.sh -E $<TARGET_FILE:bootrom> -E ${NAME}/fw_payload.elf
```
engdoreis, line 91: Can the target name for fw_payload be used instead?

ziuziakowska: For the SW tests the test name and the binary name are the same, but I wasn't able to do that for OpenSBI builds... I don't think there is a way declare install artefacts for `ExternalProject` outputs. I would like to come back to the build system in future work because it is becoming a bit inflexible.

engdoreis (approved):

PR #463, opened 2026-04-23
[rdl] Fix DRAM size

raylau1:
This fixes the size of DRAM in RDL and Verilator.


marnovandermaas (approved):
This matches the DRAMLength in top_pkg.sv.

marnovandermaas:
Do you mind adding a bit of a description into your commit message? Maybe describing that we fix the DRAM size to take into account tag storage.

PR #473, opened 2026-04-27
[chip, I2C, DV] Top level I2C host test

### LINKED ISSUE #144: I2C DV - Top-level smoke tests
- [x]  Basic reg access in C
- [x] https://github.com/lowRISC/mocha/issues/507
- [ ] https://github.com/lowRISC/mocha/issues/509

### LINKED ISSUE #507: I2C DV -- Host TX-RX test
Write `SW` that configure I2C IP in host mode and transmits / receive `n` bytes to a `reactive` UVM agent

### LINKED ISSUE #508: Controller <-> Target test
Created accidentally. A mimic of #507 

Close by #507

KinzaQamar:
Commands to test: 
`dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_host_tx_rx -t xcelium -fs 1`
`dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_host_tx_rx_cheri -t xcelium -fs 1`

Close #507

### DIFF
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_base_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_base_vseq.sv
@@ -22,7 +22,7 @@ class top_chip_dv_base_vseq extends uvm_sequence;
 
   // Class specific methods
   extern function void set_handles();
-  extern task dut_init(string reset_kind = "HARD");
+  extern virtual task dut_init(string reset_kind = "HARD");
   extern task apply_reset(string kind = "HARD");
   extern task wait_for_sw_test_done();
   // Backdoor-read or override a const symbol in SW to modify the behavior of the test.
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
@@ -1,0 +1,53 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Virtual sequence for testing the DUT as an I2C host/controller.
+//
+// The agent is configured in Device mode, acting as a reactive target that responds to transactions
+// initiated by the DUT. The i2c_monitor watches the bus and forwards observed items to the
+// i2c_sequencer via an analysis port. The i2c_device_response_seq then creates ACK/NACK/Rdata
+// responses and i2c_driver drives them back to the DUT in response.
+class top_chip_dv_i2c_host_tx_rx_vseq extends top_chip_dv_i2c_tx_rx_vseq;
+  `uvm_object_utils(top_chip_dv_i2c_host_tx_rx_vseq)
+
+  extern function new(string name="");
+  extern task body();
+  extern virtual task dut_init(string reset_kind = "HARD");
+endclass : top_chip_dv_i2c_host_tx_rx_vseq
+
+function top_chip_dv_i2c_host_tx_rx_vseq::new(string name = "");
+  super.new(name);
+endfunction
+
+task top_chip_dv_i2c_host_tx_rx_vseq::dut_init(string reset_kind = "HARD");
+  super.dut_init(reset_kind);
+
+  // Read the timing parameters through SW backdoor load
+  sw_symbol_backdoor_read("sys_clk_period_ns", sw_sys_clk_period_ns);
+  sw_symbol_backdoor_read("scl_low_time_ns", sw_scl_low_time_ns);
+  sw_symbol_backdoor_read("hold_data_time_ns", sw_data_hold_time_ns);
+
+  scl_low_cycles    = round_up_divide({sw_scl_low_time_ns[1], sw_scl_low_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+  sda_hold_cycles   = round_up_divide({sw_data_hold_time_ns[1], sw_data_hold_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+
+endtask
+
+task top_chip_dv_i2c_host_tx_rx_vseq::body();
+  i2c_device_response_seq seq = i2c_device_response_seq::type_id::create("seq");
+
+  // Configure the agent to be reactive
+  cfg.m_i2c_agent_cfg.if_mode = Device;
+  super.body();
+
+  `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
+  `uvm_info(`gfn, "Starting I2C Host TX-RX test", UVM_LOW)
+
+  configure_agent_timing();
+  print_i2c_timing_cfg();
+  fork
+    seq.start(p_sequencer.i2c_sqr);
+  join_none
+endtask
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
@@ -1,0 +1,84 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class top_chip_dv_i2c_tx_rx_vseq extends top_chip_dv_base_vseq;
+  `uvm_object_utils(top_chip_dv_i2c_tx_rx_vseq)
+
+  // Below variables will get assigned through SW backdoor load. They are defined as byte size
+  // arrays because "sw_symbol_backdoor_read/overwrite" takes an array as an argument to write or
+  // read the SW symbol.
+  protected bit [7:0] sw_sys_clk_period_ns[1];
+  protected bit [7:0] sw_scl_low_time_ns[2];
+  protected bit [7:0] sw_data_hold_time_ns[2];
+
+  // The timing parameters in cycles used by the agent to add relevant delays before driving the
+  // responses.
+  protected bit [15:0] scl_low_cycles;
+  protected bit [15:0] sda_hold_cycles;
+
+  extern function new(string name="");
+
+  // Returns the ceiling of (a / b), converting a timing parameter "a" in nanoseconds to an integer
+  // number of cycles by rounding up.
+  extern protected function int unsigned round_up_divide(int unsigned a, int unsigned b);
+
+  // Compute timing parameters utilized by the agent to add delays to the responses
+  extern protected function void configure_agent_timing();
+  extern protected function void print_i2c_timing_cfg();
+endclass : top_chip_dv_i2c_tx_rx_vseq
+
+function top_chip_dv_i2c_tx_rx_vseq::new(string name = "");
+  super.new(name);
+endfunction
+
+function int unsigned top_chip_dv_i2c_tx_rx_vseq::round_up_divide(int unsigned a, int unsigned b);
+  return (((a - 1) / b) + 1);
+endfunction
+
+function void top_chip_dv_i2c_tx_rx_vseq::configure_agent_timing();
+  // tSetupBit are the clk_i cycles before SCL goes high to drive SDA. Agent should drive SDA at
+  // least two cycles before SCL goes high.
+  int unsigned tSetupBit = 2;
+  cfg.m_i2c_agent_cfg.timing_cfg.tSetupBit = tSetupBit;
+
+  // tHoldBit are the clk_i cycles to hold SDA after SCL goes low.
+  cfg.m_i2c_agent_cfg.timing_cfg.tHoldBit = sda_hold_cycles;
+
+  // Used by i2c_if to stretch SCL by the amount of tClockpulse clk_i cycles before driving SDA. If
+  // tClockPulse is greater than the clk_i cycles taken by an SCL pulse, then i2c_monitor
+  // acknowledges the "ack" late and then drives Rdata on SDA when SCL is high.
+  cfg.m_i2c_agent_cfg.timing_cfg.tClockPulse = scl_low_cycles;
+
+  // tClockLow are the clk_i cycles that the i2c_driver use before driving SDA after stretching SCL
+  // by tClockPulse clk_i cycles. Drive SDA at least tSetupBit cycles earlier to avoid the chances
+  // of SDA interference.
+  cfg.m_i2c_agent_cfg.timing_cfg.tClockLow = scl_low_cycles - tSetupBit;
+endfunction
+
+function void top_chip_dv_i2c_tx_rx_vseq::print_i2c_timing_cfg();
+  timing_cfg_t timing_cfg = cfg.m_i2c_agent_cfg.timing_cfg;
+  string str = "";
+
+  // Print the timing parameters in a tabular form
+  str = {str, "\n+----------------------+---------+"};
+  str = {str, $sformatf("\n| %-20s | %-7s |", "Timing Parameter", "Value")};
+  str = {str, "\n+----------------------+---------+"};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSetupStart", timing_cfg.tSetupStart)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tHoldStart", timing_cfg.tHoldStart)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tClockStart", timing_cfg.tClockStart)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tClockLow", timing_cfg.tClockLow)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSetupBit", timing_cfg.tSetupBit)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tClockPulse", timing_cfg.tClockPulse)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tHoldBit", timing_cfg.tHoldBit)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tClockStop", timing_cfg.tClockStop)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSetupStop", timing_cfg.tSetupStop)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tHoldStop", timing_cfg.tHoldStop)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tTimeOut", timing_cfg.tTimeOut)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "enbTimeOut", timing_cfg.enbTimeOut)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tStretchHostClock",timing_cfg.tStretchHostClock)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSdaUnstable", timing_cfg.tSdaUnstable)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSdaInterference", timing_cfg.tSdaInterference)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSclInterference", timing_cfg.tSclInterference)};
+  `uvm_info(`gfn, $sformatf("%s", str), UVM_MEDIUM);
+endfunction
--- a/hw/top_chip/dv/env/top_chip_dv_env.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_env.sv
@@ -13,6 +13,7 @@ class top_chip_dv_env extends uvm_env;
 
   // Agents
   uart_agent m_uart_agent;
+  i2c_agent  m_i2c_agent;
 
   // Standard SV/UVM methods
   extern function new(string name = "", uvm_component parent = null);
@@ -66,6 +67,10 @@ function void top_chip_dv_env::build_phase(uvm_phase phase);
   if (!uvm_config_db#(virtual clk_rst_if)::get(this, "", "sys_clk_if", cfg.sys_clk_vif)) begin
     `uvm_fatal(`gfn, "Cannot get sys_clk_vif")
   end
+
+  // Instantiate I2C agent
+  m_i2c_agent = i2c_agent::type_id::create("m_i2c_agent", this);
+  uvm_config_db#(i2c_agent_cfg)::set(this, "m_i2c_agent", "cfg", cfg.m_i2c_agent_cfg);
 
   // Instantiate UART agent
   m_uart_agent = uart_agent::type_id::create("m_uart_agent", this);
--- a/hw/top_chip/dv/env/top_chip_dv_env.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_env.sv
@@ -88,6 +88,7 @@ function void top_chip_dv_env::connect_phase(uvm_phase phase);
   // Track specific agent sequencers in the virtual sequencer.
   // Allows virtual sequences to use the agents to drive RX items.
   top_vsqr.uart_sqr = m_uart_agent.sequencer;
+  top_vsqr.i2c_sqr  = m_i2c_agent.sequencer;
 
   // Connect monitor output to matching FIFO in the virtual sequencer.
   // Allows virtual sequences to check TX items.
--- a/hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
@@ -18,6 +18,7 @@ class top_chip_dv_env_cfg extends uvm_object;
 
   // External interface agent configs
   rand uart_agent_cfg    m_uart_agent_cfg;
+  rand i2c_agent_cfg     m_i2c_agent_cfg;
 
   `uvm_object_utils_begin(top_chip_dv_env_cfg)
   `uvm_object_utils_end
@@ -45,6 +46,12 @@ function void top_chip_dv_env_cfg::initialize();
   // Configuration is required to perform meaningful monitoring
   m_uart_agent_cfg.en_tx_monitor = 0;
   m_uart_agent_cfg.en_rx_monitor = 0;
+
+  // Create I2C agent config object
+  m_i2c_agent_cfg = i2c_agent_cfg::type_id::create("m_i2c_agent_cfg");
+
+  // Disable functional coverage collection for chip level
+  m_i2c_agent_cfg.en_cov = 0;
 endfunction : initialize
 
 function void top_chip_dv_env_cfg::get_mem_image_files_from_plusargs();
--- a/hw/top_chip/dv/tb/tb.sv
+++ b/hw/top_chip/dv/tb/tb.sv
@@ -38,10 +38,23 @@ module tb;
   logic                                      rng_valid;
   logic [top_pkg::EntropySrcRngBusWidth-1:0] rng_bits;
 
+  // I2C connections
+  wire  scl;
+  wire  sda;
+  logic scl_en_o;
+  logic sda_en_o;
+  logic scl_o;
+  logic sda_o;
+
   // ------ Interfaces ------
   clk_rst_if sys_clk_if(.clk(clk), .rst_n(rst_n));
   uart_if uart_if();
   pins_if #(NUM_GPIOS) gpio_pins_if (.pins(gpio_pads));
+  i2c_if i2c_if (.clk_i(clk    ),
+                 .rst_ni(rst_n ),
+                 .scl_io(scl   ),
+                 .sda_io(sda   )
+                );
 
   // ------ Mock DRAM ------
   top_pkg::axi_dram_req_t  dram_req;
@@ -82,6 +95,13 @@ module tb;
     // UART receive and transmit.
     .uart_rx_i            (uart_if.uart_rx  ),
     .uart_tx_o            (uart_if.uart_tx  ),
+    // I2C controller/target bidirectional interface.
+    .i2c_scl_i            (scl              ),
+    .i2c_scl_o            (scl_o            ),
+    .i2c_scl_en_o         (scl_en_o         ),
+    .i2c_sda_i            (sda              ),
+    .i2c_sda_o            (sda_o            ),
+    .i2c_sda_en_o         (sda_en_o         ),
     // External Mailbox port
     .axi_mailbox_req_i    ('0               ),
     .axi_mailbox_resp_o   (                 ),
@@ -125,6 +145,10 @@ module tb;
   for (genvar i = 0; i < NUM_GPIOS; i++) begin : gen_gpio_pads
     assign gpio_pads[i] = dut_gpio_en_o[i] ? dut_gpio_o[i] : 1'bz;
   end
+
+  // Modelling the open-drain circuit
+  assign (strong0, weak1) scl = (scl_en_o) ? scl_o : 1'b1;
+  assign (strong0, weak1) sda = (sda_en_o) ? sda_o : 1'b1;
 
   // Signals to connect the sink
   logic               sim_sram_clk;
--- a/hw/top_chip/dv/tb/tb.sv
+++ b/hw/top_chip/dv/tb/tb.sv
@@ -284,6 +284,7 @@ module tb;
     uvm_config_db#(virtual clk_rst_if)::set(null, "*", "sys_clk_if", sys_clk_if);
     uvm_config_db#(virtual uart_if)::set(null, "*.env.m_uart_agent*", "vif", uart_if);
     uvm_config_db#(virtual pins_if #(NUM_GPIOS))::set(null, "*.env", "gpio_vif", gpio_pins_if);
+    uvm_config_db#(virtual i2c_if)::set(null, "*.env.m_i2c_agent", "vif", i2c_if);
 
     // SW logger and test status interfaces.
     uvm_config_db#(virtual sw_test_status_if)::set(
--- a/hw/top_chip/dv/top_chip_sim_cfg.hjson
+++ b/hw/top_chip/dv/top_chip_sim_cfg.hjson
@@ -151,7 +151,14 @@
       sw_images: ["i2c_host_tx_rx_test_vanilla_sram:5" "bootrom:5"]
       run_opts: ["+ChipMemSRAM_image_file={run_dir}/i2c_host_tx_rx_test_vanilla_sram.vmem",
                  "+ChipMemROM_image_file={run_dir}/bootrom_scrambled.vmem"]
-      }
+    }
+    {
+      name: i2c_host_tx_rx_cheri
+      uvm_test_seq: top_chip_dv_i2c_host_tx_rx_vseq
+      sw_images: ["i2c_host_tx_rx_test_cheri_sram:5" "bootrom:5"]
+      run_opts: ["+ChipMemSRAM_image_file={run_dir}/i2c_host_tx_rx_test_cheri_sram.vmem",
+                 "+ChipMemROM_image_file={run_dir}/bootrom_scrambled.vmem"]
+    }
     {
       name: i2c_device_tx_rx
       uvm_test_seq: top_chip_dv_i2c_device_tx_rx_vseq
@@ -345,6 +352,8 @@
         "spi_device_smoke_cheri",
         "spi_host_smoke",
         "spi_host_smoke_cheri",
+        "i2c_host_tx_rx",
+        "i2c_host_tx_rx_cheri",
         "gpio_smoke",
         "gpio_smoke_cheri",
         "mailbox_smoke",
@@ -404,6 +413,7 @@
       name: i2c
       tests: [
         "i2c_host_tx_rx",
+        "i2c_host_tx_rx_cheri",
         "i2c_device_tx_rx"
       ]
     }
--- a/sw/device/tests/CMakeLists.txt
+++ b/sw/device/tests/CMakeLists.txt
@@ -17,6 +17,7 @@ mocha_add_test(NAME gpio_reg_access_test SOURCES gpio/reg_access_test.c LIBRARIE
 # driven externally
 mocha_add_test(NAME gpio_smoketest SOURCES gpio/smoketest.c LIBRARIES ${LIBS} SKIP_VERILATOR SKIP_FPGA)
 mocha_add_test(NAME i2c_temperature_sensor_test SOURCES i2c/temperature_sensor_test.c LIBRARIES ${LIBS})
+mocha_add_test(NAME i2c_host_tx_rx_test SOURCES i2c/host_tx_rx_test.c LIBRARIES ${LIBS} SKIP_VERILATOR SKIP_FPGA)
 mocha_add_test(NAME mailbox_smoketest SOURCES mailbox/smoketest.c LIBRARIES ${LIBS})
 mocha_add_test(NAME plic_smoketest SOURCES plic/smoketest.c LIBRARIES ${LIBS})
 mocha_add_test(NAME pwrmgr_smoketest SOURCES pwrmgr/smoketest.c LIBRARIES ${LIBS})
--- a/sw/device/tests/i2c/host_tx_rx_test.c
+++ b/sw/device/tests/i2c/host_tx_rx_test.c
@@ -8,6 +8,15 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+// The const variables below are treated as symbols read by top_chip_dv_i2c_host_tx_rx_vseq in order
+// to calculate agent timing parameters.
+const uint8_t sys_clk_period_ns = SYSCLK_NS;
+
+// The constants are the spec minimums for standard mode speed except hold_data_time_ns which should
+// be at least one according to OpenTitan's programming guide.
+const uint16_t scl_low_time_ns = 4700;
+const uint16_t hold_data_time_ns = 1;
+
 enum : uint8_t {
     device_addr = 0x3A,
     num_bytes = 0x8,
--- a/sw/device/tests/i2c/host_tx_rx_test.c
+++ b/sw/device/tests/i2c/host_tx_rx_test.c
@@ -1,0 +1,67 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "hal/i2c.h"
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+enum : uint8_t {
+    device_addr = 0x3A,
+    num_bytes = 0x8,
+};
+
+static bool write_transfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_bytes)
+{
+    // Start a write transfer
+    i2c_write_bytes(i2c, addr, data, num_bytes);
+    return i2c_wait_transfer_finish(i2c);
+}
+
+static bool read_transfer(i2c_t i2c, uint8_t addr, uint8_t num_bytes)
+{
+    // Start a read transfer
+    i2c_read_bytes(i2c, addr, num_bytes);
+    return i2c_wait_transfer_finish(i2c);
+}
+
+static bool drive_transfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_bytes)
+{
+    bool write_transfer_status = write_transfer(i2c, addr, data, num_bytes);
+    bool read_transfer_status = read_transfer(i2c, addr, num_bytes);
+
+    return (write_transfer_status && read_transfer_status);
+}
+
+static bool host_tx_rx_test(i2c_t i2c)
+{
+    // Data bytes to send to the target's receiver.
+    uint8_t data_bytes[num_bytes];
+
+    // Write walking 1's pattern
+    for (uint8_t i = 0; i < num_bytes; i++) {
+        data_bytes[i] = 1u << (i % 8);
+    }
+
+    if (!drive_transfer(i2c, device_addr, data_bytes, num_bytes)) {
+        return false;
+    }
+
+    for (uint8_t i = 0; i < num_bytes; i++) {
+        if (data_bytes[i] != i2c_rdata_byte(i2c)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool test_main()
+{
+    i2c_t i2c = mocha_system_i2c();
+    i2c_init(i2c, i2c_speed_mode_standard);
+    i2c_enable_controller_mode(i2c);
+    return host_tx_rx_test(i2c);
+}
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_vseq_list.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_vseq_list.sv
@@ -5,3 +5,5 @@
 `include "top_chip_dv_base_vseq.sv"
 `include "top_chip_dv_uart_base_vseq.sv"
 `include "top_chip_dv_gpio_smoke_vseq.sv"
+`include "top_chip_dv_i2c_tx_rx_vseq.sv"
+`include "top_chip_dv_i2c_host_tx_rx_vseq.sv"
--- a/hw/top_chip/dv/env/top_chip_dv_env.core
+++ b/hw/top_chip/dv/env/top_chip_dv_env.core
@@ -13,6 +13,7 @@ filesets:
       - lowrisc:dv:uart_agent
       - lowrisc:dv:common_ifs
       - lowrisc:mocha_dv:gpio_env
+      - lowrisc:dv:i2c_env
     files:
       - top_chip_dv_env_pkg.sv
       - mem_clear_util.sv: {is_include_file: true}
--- a/hw/top_chip/dv/env/top_chip_dv_env.core
+++ b/hw/top_chip/dv/env/top_chip_dv_env.core
@@ -25,6 +25,8 @@ filesets:
       - seq_lib/top_chip_dv_base_vseq.sv: {is_include_file: true}
       - seq_lib/top_chip_dv_uart_base_vseq.sv: {is_include_file: true}
       - seq_lib/top_chip_dv_gpio_smoke_vseq.sv: {is_include_file: true}
+      - seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv: {is_include_file: true}
+      - seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv: {is_include_file: true}
     file_type: systemVerilogSource
 
 targets:
--- a/hw/top_chip/dv/env/top_chip_dv_virtual_sequencer.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_virtual_sequencer.sv
@@ -14,6 +14,7 @@ class top_chip_dv_virtual_sequencer extends uvm_sequencer;
   // Handles to specific interface agent sequencers. Used by some virtual
   // sequences to drive RX (to-chip) items.
   uart_sequencer uart_sqr;
+  i2c_sequencer  i2c_sqr;
 
   // FIFOs for monitor output. Used by some virtual sequences to check
   // TX (from-chip) items.


KinzaQamar:
DV changes are open for review now. 

SW changes will be reviewable once #565 and #535 are merged

hw/top_chip/dv/env/seq_lib/top_chip_dv_base_vseq.sv
```diff
@@ -22,7 +22,7 @@ class top_chip_dv_base_vseq extends uvm_sequence;
 
   // Class specific methods
   extern function void set_handles();
-  extern task dut_init(string reset_kind = "HARD");
+  extern virtual task dut_init(string reset_kind = "HARD");
```
elliotb-lowrisc, line 25: Forgive my ignorance, but why has `virtual` been added here?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,84 @@
...
+
+  // Used by i2c_if to stretch SCL by tClockpulse before driving SDA. If tClockPulse is greater
+  // than scl_low_cycles then i2c_monitor acknowledge the Ack later and then drives Rdata when SCL
+  // is high.
+  cfg.m_i2c_agent_cfg.timing_cfg.tClockPulse = scl_low_cycles;
+
+  // tClockLow are the SCL low cycles that the i2c_driver use before driving SDA after stretching
+  // SCL by tClockPulse cycles. Drive SDA atleast tSetBit cycles earlier to avoid the chances of
+  // SDA interference.
+  cfg.m_i2c_agent_cfg.timing_cfg.tClockLow = scl_low_cycles - tSetupBit;
```
elliotb-lowrisc: Do you not need to set any of the other timing parameters used by i2c_if?

hw/top_chip/dv/env/seq_lib/top_chip_dv_base_vseq.sv
KinzaQamar: `top_chip_dv_tx_rx_vseq` extends from this class. I wanted to use `dut_init()` function to initialized timing parameters through `sw_symbol_backdoor_read`. If a subclass overrides a method of a parent class without the virtual keyword then the method of parent class will be called at runtime.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: Not for the `host` test. The agent can send `ack`, `nack` and `Rdata` to the `host` I2C block. `i2c_driver` will use [drive_device_item](https://github.com/lowRISC/opentitan/blob/d71121d8452ff8797abf949219a67726c94ea8eb/hw/dv/sv/i2c_agent/i2c_driver.sv#L157-L197) function to send the responses which then call `device_send_bit` in `i2c_if`. All these functions use only the parameters set in this `vseq`.

There are other parameters like `tSdaUnstable` or `tTimeout` that could be used but they are not relevant for this test.

elliotb-lowrisc: Thanks for explaining

hw/top_chip/dv/env/seq_lib/top_chip_dv_base_vseq.sv
elliotb-lowrisc: I see, but then why call `super.dut_init(reset_kind);` in the subclass task if it is overriding the parent task?

KinzaQamar: So, the actual definition of the function in the parent class is:
```
task top_chip_dv_base_vseq::dut_init(string reset_kind = "HARD");
  if (do_apply_reset) begin
    apply_reset(reset_kind);
  end
endtask : dut_init
```
This function is called in the `pre_start()` UVM phase. The subclass extended the functionality but also wants to grab the existing functionality from the parent class. Without `super.dut_init()` the function in the sub class will only do:

```
task top_chip_dv_i2c_host_tx_rx_vseq::dut_init(string reset_kind = "HARD");
  // Read the timing parameters through SW backdoor load
  sw_symbol_backdoor_read("SysClkPeriodNS", sw_sys_clk_period_ns);
  sw_symbol_backdoor_read("SCLLowTimeNs", sw_scl_low_time_ns);
  sw_symbol_backdoor_read("HoldDataTimeNs", sw_data_hold_time_ns);

  scl_period_cycles = round_up_divide({sw_scl_clk_period_ns[1], sw_scl_clk_period_ns[0]},
                                      sw_sys_clk_period_ns[0]);
  scl_low_cycles    = round_up_divide({sw_scl_low_time_ns[1], sw_scl_low_time_ns[0]},
                                      sw_sys_clk_period_ns[0]);
  sda_hold_cycles   = round_up_divide({sw_data_hold_time_ns[1], sw_data_hold_time_ns[0]},
                                      sw_sys_clk_period_ns[0]);

endtask
```

With super():
```
task top_chip_dv_i2c_host_tx_rx_vseq::dut_init(string reset_kind = "HARD");
  // Read the timing parameters through SW backdoor load
  sw_symbol_backdoor_read("SysClkPeriodNS", sw_sys_clk_period_ns);
  sw_symbol_backdoor_read("SCLLowTimeNs", sw_scl_low_time_ns);
  sw_symbol_backdoor_read("HoldDataTimeNs", sw_data_hold_time_ns);

  scl_period_cycles = round_up_divide({sw_scl_clk_period_ns[1], sw_scl_clk_period_ns[0]},
                                      sw_sys_clk_period_ns[0]);
  scl_low_cycles    = round_up_divide({sw_scl_low_time_ns[1], sw_scl_low_time_ns[0]},
                                      sw_sys_clk_period_ns[0]);
  sda_hold_cycles   = round_up_divide({sw_data_hold_time_ns[1], sw_data_hold_time_ns[0]},
                                      sw_sys_clk_period_ns[0]);
  if (do_apply_reset) begin
    apply_reset(reset_kind);
  end
endtask
```

Without `virtual`, the `dut_init()` of the base_class will be executed in the `pre_start()` phase

elliotb-lowrisc: Ah, thanks for explaining. Sorry for my slowness in understanding.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
```diff
@@ -0,0 +1,66 @@
+// Copyright lowRISC contributors (OpenTitan project).
```
martin-velay:
```suggestion
// Copyright lowRISC contributors (COSMIC project).
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
```diff
@@ -0,0 +1,66 @@
...
+  str = {str, $sformatf("\n    timing_cfg.enbTimeOut        : %d",
+            cfg.m_i2c_agent_cfg.timing_cfg.enbTimeOut)};
+  str = {str, $sformatf("\n    timing_cfg.tStretchHostClock : %d",
+            cfg.m_i2c_agent_cfg.timing_cfg.tStretchHostClock)};
+  str = {str, $sformatf("\n    timing_cfg.tSdaUnstable      : %d",
+            cfg.m_i2c_agent_cfg.timing_cfg.tSdaUnstable)};
+  str = {str, $sformatf("\n    timing_cfg.tSdaInterference  : %d",
+            cfg.m_i2c_agent_cfg.timing_cfg.tSdaInterference)};
+  str = {str, $sformatf("\n    timing_cfg.tSclInterference  : %d",
+            cfg.m_i2c_agent_cfg.timing_cfg.tSclInterference)};
```
martin-velay: Can you do the following, as we have the right to go up to 120 characters per line in Mocha:
```suggestion
  str = {str, $sformatf("\n  timing_cfg.tSetupStart       : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSetupStart)};
  str = {str, $sformatf("\n  timing_cfg.tHoldStart        : %d", cfg.m_i2c_agent_cfg.timing_cfg.tHoldStart)};
  str = {str, $sformatf("\n  timing_cfg.tClockStart       : %d", cfg.m_i2c_agent_cfg.timing_cfg.tClockStart)};
  str = {str, $sformatf("\n  timing_cfg.tClockLow         : %d", cfg.m_i2c_agent_cfg.timing_cfg.tClockLow)};
  str = {str, $sformatf("\n  timing_cfg.tSetupBit         : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSetupBit)};
  str = {str, $sformatf("\n  timing_cfg.tClockPulse       : %d", cfg.m_i2c_agent_cfg.timing_cfg.tClockPulse)};
  str = {str, $sformatf("\n  timing_cfg.tHoldBit          : %d", cfg.m_i2c_agent_cfg.timing_cfg.tHoldBit)};
  str = {str, $sformatf("\n  timing_cfg.tClockStop        : %d", cfg.m_i2c_agent_cfg.timing_cfg.tClockStop)};
  str = {str, $sformatf("\n  timing_cfg.tSetupStop        : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSetupStop)};
  str = {str, $sformatf("\n  timing_cfg.tHoldStop         : %d", cfg.m_i2c_agent_cfg.timing_cfg.tHoldStop)};
  str = {str, $sformatf("\n  timing_cfg.tTimeOut          : %d", cfg.m_i2c_agent_cfg.timing_cfg.tTimeOut)};
  str = {str, $sformatf("\n  timing_cfg.enbTimeOut        : %d", cfg.m_i2c_agent_cfg.timing_cfg.enbTimeOut)};
  str = {str, $sformatf("\n  timing_cfg.tStretchHostClock : %d", cfg.m_i2c_agent_cfg.timing_cfg.tStretchHostClock)};
  str = {str, $sformatf("\n  timing_cfg.tSdaUnstable      : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSdaUnstable)};
  str = {str, $sformatf("\n  timing_cfg.tSdaInterference  : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSdaInterference)};
  str = {str, $sformatf("\n  timing_cfg.tSclInterference  : %d", cfg.m_i2c_agent_cfg.timing_cfg.tSclInterference)};
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,83 @@
...
+
+function top_chip_dv_i2c_host_tx_rx_vseq::new(string name = "");
+  super.new(name);
+endfunction : new
+
+task top_chip_dv_i2c_host_tx_rx_vseq::dut_init(string reset_kind = "HARD");
+  // Read the timing parameters through SW backdoor load
+  sw_symbol_backdoor_read("SysClkPeriodNS", sw_sys_clk_period_ns);
+  sw_symbol_backdoor_read("SCLLowTimeNs", sw_scl_low_time_ns);
+  sw_symbol_backdoor_read("HoldDataTimeNs", sw_data_hold_time_ns);
```
martin-velay: Where `sw_scl_clk_period_ns` is populated? I think something like that is missing?
`sw_symbol_backdoor_read("SCLClkPeriodNS", sw_scl_clk_period_ns);`

sw/device/lib/hal/i2c.c
```diff
@@ -22,36 +31,105 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
     return (uint16_t)result;
 }
 
-void i2c_init(i2c_t i2c)
+uint16_t i2c_calc_scl_high_cycles(uint16_t rise_cycles, uint16_t fall_cycles,
```
martin-velay, line 25: Maybe better to compute in `uint32_t`,  in case the sum exceeds 16 bits. Then maybe check if it's not going over 16 bits and return in 16 bits?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,83 @@
...
+  // The timing parameter in cycles used by the agent to add relevant delays before driving Ack,
+  // Nack and Rdata
+  local bit [15:0] scl_period_cycles;
+  local bit [15:0] scl_low_cycles;
+  local bit [15:0] sda_hold_cycles;
+
+  extern function new(string name="");
+  extern task body();
+  extern virtual task dut_init(string reset_kind = "HARD");
+  extern local function void configure_agent_timing();
```
martin-velay: Do we really need to restrict this function to be `local`?

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
```diff
@@ -18,6 +18,7 @@ class top_chip_dv_env_cfg extends uvm_object;
 
   // External interface agent configs
   rand uart_agent_cfg    m_uart_agent_cfg;
+  i2c_agent_cfg          m_i2c_agent_cfg;
```
martin-velay:
```suggestion
  rand i2c_agent_cfg     m_i2c_agent_cfg;
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,83 @@
...
+  sw_symbol_backdoor_read("HoldDataTimeNs", sw_data_hold_time_ns);
+
+  scl_period_cycles = round_up_divide({sw_scl_clk_period_ns[1], sw_scl_clk_period_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+  scl_low_cycles    = round_up_divide({sw_scl_low_time_ns[1], sw_scl_low_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+  sda_hold_cycles   = round_up_divide({sw_data_hold_time_ns[1], sw_data_hold_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+
+  super.dut_init(reset_kind);
```
martin-velay: Why do we call the `super` afterward? If there is no reason, could you move it at the beginning of this task?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,83 @@
...
+endtask
+
+function void top_chip_dv_i2c_host_tx_rx_vseq::configure_agent_timing();
+  // tSetupBit are the cycles before SCL goes high to drive SDA. Agent should drive atleast two
```
martin-velay:
```suggestion
  // tSetupBit are the cycles before SCL goes high to drive SDA. Agent should drive at least two
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,83 @@
...
+  // tHoldBit are the cycles to hold SDA after SCL goes low.
+  cfg.m_i2c_agent_cfg.timing_cfg.tHoldBit = sda_hold_cycles;
+
+  // Used by i2c_if to stretch SCL by tClockpulse before driving SDA. If tClockPulse is greater
+  // than scl_low_cycles then i2c_monitor acknowledge the Ack later and then drives Rdata when SCL
+  // is high.
+  cfg.m_i2c_agent_cfg.timing_cfg.tClockPulse = scl_low_cycles;
+
+  // tClockLow are the SCL low cycles that the i2c_driver use before driving SDA after stretching
+  // SCL by tClockPulse cycles. Drive SDA atleast tSetBit cycles earlier to avoid the chances of
```
martin-velay:
```suggestion
  // SCL by tClockPulse cycles. Drive SDA at least tSetBit cycles earlier to avoid the chances of
```

sw/device/lib/hal/i2c.c
```diff
@@ -22,36 +31,105 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
...
+                                  uint16_t scl_period_cycles, uint16_t scl_low_cycles,
+                                  uint16_t scl_high_cycles_min)
 {
-    // -- Set timing parameters --
-    //
-    // Using Standard-mode (100 kbits/s) constants, taken from the NXP I^2C specification
-    // "UM10204" Table 10 (rev. 6) / Table 11 (rev. 7).
-    // Faster modes will require different/adjustable constants and checking that SCL high/low
-    // cycles calculated are sufficient to still allow the clock stretching logic to function.
+    // scl_high_time should be atleast 4 cycles to aid correct clock streching
```
martin-velay:
```suggestion
    // scl_high_time should be at least 4 cycles to aid correct clock stretching
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
```diff
@@ -0,0 +1,66 @@
...
+  // because "sw_symbol_backdoor_read/overwrite" takes an array as an argument to write or read the
+  // SW symbol.
+  protected bit [7:0] sw_sys_clk_period_ns[1];
+  protected bit [7:0] sw_scl_clk_period_ns[2];
+  protected bit [7:0] sw_scl_low_time_ns[2];
+  protected bit [7:0] sw_data_hold_time_ns[2];
+
+  extern function new(string name="");
+
+  // Obtain an integer minimum of timing parameter "a" and round up to the next highest integer.
```
martin-velay: Could you reformulate, I feel that the word minimum is confusing, but it might be my comprehension

sw/device/lib/hal/i2c.c
```diff
@@ -22,36 +31,105 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
...
+
+    return scl_high_cycles;
+}
+
+// Calculate the minimum allowable value for each timing parameter taken from the NXP I^2C
+// specification "UM10204" Table 10 (rev. 6) / Table 11 (rev. 7).
+//
+// The values for Rise and Fall times for Fast mode are taken as spec minimum. For Fast plus mode,
+// the values are taken from OT's i2c_host_tx_rx_test.c test.
+i2c_timing_params_t compute_minimum_timing_paramaters(i2c_speed_mode_t speed)
```
martin-velay: Nit:
```suggestion
i2c_timing_params_t compute_minimum_timing_parameters(i2c_speed_mode_t speed)
```

Also, this function shouldn't be in the header file instead?

sw/device/lib/hal/i2c.c
```diff
@@ -22,36 +31,105 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
...
+        (scl_high_cycles > scl_high_cycles_min) ? scl_high_cycles : scl_high_cycles_min;
+
+    return scl_high_cycles;
+}
+
+// Calculate the minimum allowable value for each timing parameter taken from the NXP I^2C
+// specification "UM10204" Table 10 (rev. 6) / Table 11 (rev. 7).
+//
+// The values for Rise and Fall times for Fast mode are taken as spec minimum. For Fast plus mode,
+// the values are taken from OT's i2c_host_tx_rx_test.c test.
```
martin-velay:
```suggestion
// the values are taken from OpenTitan's i2c_host_tx_rx_test.c test.
```

sw/device/lib/hal/i2c.h
```diff
@@ -88,7 +88,29 @@
 #define I2C_RISE_NS (450)
 #define I2C_FALL_NS (120)
 
-void i2c_init(i2c_t i2c);
+// All the speed modes supported by OT's I2C block
```
martin-velay, line 91:
```suggestion
// All the speed modes supported by OpenTitan's I2C block
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: This is already there on L26

sw/device/lib/hal/i2c.c
KinzaQamar: You are suggesting the same thing?

Only function declarations are present in `i2c.h` [here](https://github.com/lowRISC/mocha/blob/main/sw/device/lib/hal/i2c.h)

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
KinzaQamar: I think you are right. I meant spec minimum

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: I will probably move it to the base vseq (`top_chip_dv_i2c_tx_rx_vseq`) and make it `protected`

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: No reason but just a preference to override the function first and then call the parent class definition later

sw/device/lib/hal/i2c.c
KinzaQamar: I think your comment is about L51: 
`uint16_t scl_high_cycles = scl_period_cycles - (scl_low_cycles + rise_cycles + fall_cycles);`

thinking about it `scl_period_cycles` should satisfy the equation:
`scl_period_cycles = rise_cycles + fall_cycles + high_cycle + low_cycles`

`scl_period_cycles` itself is 16-bits so  expression `scl_low_cycles + rise_cycles + fall_cycles` can't go out of bound 16bits I think

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
KinzaQamar: There is no member in `i2c_agent_cfg` that is declared with a `rand` keyword

KinzaQamar:
Refer to the note about why `host_test.c` is failing on verilator[507#issuecomment-4569471032)](https://github.com/lowRISC/mocha/issues/507#issuecomment-4569471032). I've suggest few options but I need few opinions as well.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
elliotb-lowrisc: Looks like L26 is about the sys clock, not the scl clock.

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
elliotb-lowrisc: If I understand correctly, the point of adding `rand` is so that the whole cfg object instance is able to go through constrained randomisation

KinzaQamar: My understanding is that if you have a class member declared with `rand` then this will work.  Only the variables that are declared with `rand` will go through `randomization`. The class constructor initialize the non-rand variable with their default values. You can randomize non-rand variables through `std::randomize` explicitly.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: Oops, in which case I am sorry @martin-velay for acting blind.

I forgot to through it away. This is not in use

Thanks @elliotb-lowrisc

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
elliotb-lowrisc: You're right, I just tried printing the value of `m_i2c_agent_cfg.target_addr0` after it is created, and it comes back as zero every time.

KinzaQamar:
> Refer to the note about why `host_test.c` is failing on verilator[507#issuecomment-4569471032)](https://github.com/lowRISC/mocha/issues/507#issuecomment-4569471032). I've suggest few options but I need few opinions as well.

@marnovandermaas suggested to skip verilator and FPGA images for `host_test`

sw/device/lib/hal/i2c.c
elliotb-lowrisc: GitHub isn't highlighting it well, but "paramaters" is indeed spelt incorrectly here.

Correct me if I'm wrong, but I believe that this function is only used by the init() function, so is fine to be omitted from the header file. You could also add `static` to tell the compiler (linker?) that is is local-only I believe, but it's not needed as far as I'm aware.

sw/device/lib/hal/i2c.c
elliotb-lowrisc: The part where I might expect an overflow is `scl_low_cycles + rise_cycles + fall_cycles`, before the subtraction.

To check whether this is OK, I suggest you calculate the maximum clock speed (smallest `SYSCLK_NS`) that could cause the above to overflow with standard mode timing. If it's stupidly fast (e.g. 100 GHz), then it's probably safe.

sw/device/lib/hal/i2c.c
martin-velay: Yes, `param**e**ters` instead of `param**a**ters`

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "hal/i2c.h"
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+bool i2c_host_wr_xfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_wr_bytes)
```
elliotb-lowrisc: No need for the "i2c_host" prefix here or on the read version

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,68 @@
...
+    if (host_rx_tx_test(i2c, 0x48u, wr_data_bytes, sizeof(wr_data_bytes))) {
+        uint8_t rd_data_bytes[sizeof(wr_data_bytes)];
+        for (uint8_t i = 0; i < (uint8_t)(sizeof(rd_data_bytes)); i++) {
+            rd_data_bytes[i] = i2c_rdata_byte(i2c);
+            if (wr_data_bytes[i] != rd_data_bytes[i]) {
+                return false;
+            }
+        }
+        success = true;
+    }
```
elliotb-lowrisc: I think this would be better outside of test_main. Generally only the peripheral setup is in the test_main so that the test can be reused in a larger test suite more easily (I think?)

sw/device/tests/i2c/host_tx_rx_test.c
elliotb-lowrisc, line 1: I think we should rename this and the existing smoketest.c to better capture what they do. In this case, it is testing reading and writing to a memory-like I^2C device. The existing "smoketest" does similar, only to a temperature sensor I^2C device. I think having test names that capture whether the i2c block is operating in host/device mode and what it is attempting to communicate with would be much clearer.

elliotb-lowrisc (commented):
Generally looking decent to me. I have a few comments on making the SW test a bit clearer and in keeping with other tests

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
martin-velay: Well, for the moment this works as the memories are loaded at compile time, but I think this call should be moved before as it calls `apply_reset` and if this task is extended to implement some kind of fault injection in the SRAM for example. Then you may end-up having different things than what you'd have read before the reset.

In my experience, in general it's better to call the `super` first and then do your own things

hw/top_chip/dv/tb/tb.sv
```diff
@@ -38,10 +38,22 @@ module tb;
...
+  logic sda_o;
+
   // ------ Interfaces ------
   clk_rst_if sys_clk_if(.clk(clk), .rst_n(rst_n));
   uart_if uart_if();
   pins_if #(NUM_GPIOS) gpio_pins_if (.pins(gpio_pads));
+  i2c_if i2c_if (.clk_i(clk),
+                 .rst_ni(rst_n),
+                 .scl_io(scl),
+                 .sda_io(sda));
```
martin-velay: Nit: to be coherent with the rest of the code
```suggestion
  i2c_if i2c_if (
    .clk_i  (clk    ),
    .rst_ni (rst_n  ),
    .scl_io (scl    ),
    .sda_io (sda    )
  );
```

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
martin-velay: It is a common (and a good) practice to declare this config class as `rand` even though there is no `rand` variables inside. If this becomes the case then you are already ready for it and no change is required.

Often config classes have randomized variables in order to explore more unexpected configurations and discover more bugs.

sw/device/tests/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ mocha_add_test(NAME gpio_reg_access_test SOURCES gpio/reg_access_test.c LIBRARIE
 # driven externally
 mocha_add_test(NAME gpio_smoketest SOURCES gpio/smoketest.c LIBRARIES ${LIBS} SKIP_VERILATOR SKIP_FPGA)
 mocha_add_test(NAME i2c_smoketest SOURCES i2c/smoketest.c LIBRARIES ${LIBS})
+mocha_add_test(NAME i2c_host_tx_rx_test SOURCES i2c/host_test.c LIBRARIES ${LIBS})
```
martin-velay: As you've mentioned, Marno has proposed to skip this test for Verilator and FPGA, remember to update this line:

```suggestion
mocha_add_test(NAME i2c_host_tx_rx_test SOURCES i2c/host_test.c LIBRARIES ${LIBS} SKIP_VERILATOR SKIP_FPGA)
```

hw/top_chip/dv/env/top_chip_dv_env.sv
```diff
@@ -67,6 +68,12 @@ function void top_chip_dv_env::build_phase(uvm_phase phase);
...
 
+  // Set I2C agent config object for I2C agent
+  uvm_config_db#(i2c_agent_cfg)::set(this, "m_i2c_agent", "cfg", cfg.m_i2c_agent_cfg);
+
+  // Create I2C agent
+  m_i2c_agent = i2c_agent::type_id::create("m_i2c_agent", this);
+
   // Instantiate UART agent
   m_uart_agent = uart_agent::type_id::create("m_uart_agent", this);
   uvm_config_db#(uart_agent_cfg)::set(this, "m_uart_agent*", "cfg", cfg.m_uart_agent_cfg);
```
martin-velay: Nit - for consistency:

```suggestion
  // Instantiate UART agent
  m_uart_agent = uart_agent::type_id::create("m_uart_agent", this);
  uvm_config_db#(uart_agent_cfg)::set(this, "m_uart_agent*", "cfg", cfg.m_uart_agent_cfg);

  // Instantiate I2C agent
  m_i2c_agent = i2c_agent::type_id::create("m_i2c_agent", this);
  uvm_config_db#(i2c_agent_cfg)::set(this, "m_i2c_agent", "cfg", cfg.m_i2c_agent_cfg);
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,54 @@
...
+  sw_symbol_backdoor_read("SCLLowTimeNs", sw_scl_low_time_ns);
+  sw_symbol_backdoor_read("HoldDataTimeNs", sw_data_hold_time_ns);
+
+  scl_low_cycles    = round_up_divide({sw_scl_low_time_ns[1], sw_scl_low_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+  sda_hold_cycles   = round_up_divide({sw_data_hold_time_ns[1], sw_data_hold_time_ns[0]},
+                                      sw_sys_clk_period_ns[0]);
+
+  super.dut_init(reset_kind);
+endtask
```
martin-velay, line 36: Nit - for consistency

```suggestion
endtask : dut_init
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -0,0 +1,54 @@
...
+
+task top_chip_dv_i2c_host_tx_rx_vseq::body();
```
martin-velay, line 38: You should add a call to the super in case there's something added there in the future:
`super.body();`

Also, for consistency, you can have these lines at the very beginning like we do for other vseqs:

```sv
  `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
  `uvm_info(`gfn, "Starting I2C Host TX-RX test", UVM_LOW)
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
```diff
@@ -0,0 +1,84 @@
...
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSetupStop", timing_cfg.tSetupStop)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tHoldStop", timing_cfg.tHoldStop)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tTimeOut", timing_cfg.tTimeOut)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "enbTimeOut", timing_cfg.enbTimeOut)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tStretchHostClock",timing_cfg.tStretchHostClock)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSdaUnstable", timing_cfg.tSdaUnstable)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSdaInterference", timing_cfg.tSdaInterference)};
+  str = {str, $sformatf("\n| %-20s | %7d |", "tSclInterference", timing_cfg.tSclInterference)};
+  `uvm_info(`gfn, $sformatf("%s", str), UVM_MEDIUM);
+endfunction
```
martin-velay, line 84: Nit (applies above too):

```suggestion
endfunction : print_i2c_timing_cfg
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: I prefer to remove the labels for consistency. Done now. Thanks

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
KinzaQamar: Addressed here : [/changes#r3349814978](https://github.com/lowRISC/mocha/pull/473/changes#r3349814978)

sw/device/lib/hal/i2c.c
```diff
@@ -8,6 +8,14 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+// Below symbols are going to be read by top_chip_dv_i2c_host_tx_rx_vseq in order to calculate agent
```
engdoreis: This should not be in the hal, this is test specific and the should be moved there.

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    // Return status of the test
+    bool success = false;
+
+    // Data bytes to send to the target's receiver.
+    uint8_t wr_data_bytes[num_bytes];
+
+    // Write walking 1's pattern
+    for (uint8_t i = 0; i < num_bytes; i++) {
+        wr_data_bytes[i] = 1u << i;
```
engdoreis:
```suggestion
        wr_data_bytes[i] = 1u << (i % 8);
```

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    i2c_t i2c = mocha_system_i2c();
+
+    // Configure the timing registers according to the I2C speed mode
+    i2c_init(i2c, i2c_speed_mode_standard);
+
+    // -- Configure IP for Controller mode --
+    i2c_enable_controller_mode(i2c);
+
+    return host_tx_rx_test(i2c, 0x3Au, 0x8u);
```
engdoreis: Can you create a constant for this address with a suggestive name?

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+    i2c_read_bytes(i2c, addr, num_rd_bytes);
+    return i2c_wait_read_finish(i2c);
+}
+
+static bool drive_xfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_bytes)
+{
+    bool rd_xfer_status, wr_xfer_status;
+
+    wr_xfer_status = wr_xfer(i2c, addr, data, num_bytes);
+    rd_xfer_status = rd_xfer(i2c, addr, num_bytes);
```
engdoreis: Nit.
```suggestion
    bool wr_xfer_status = wr_xfer(i2c, addr, data, num_bytes);
    bool rd_xfer_status = rd_xfer(i2c, addr, num_bytes);
```

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+        // part of write transfer. Store those bytes in an array and compare it with the wr_data_bytes
+        // array.
+        uint8_t rd_data_bytes[num_bytes];
+        for (uint8_t i = 0; i < num_bytes; i++) {
+            rd_data_bytes[i] = i2c_rdata_byte(i2c);
+            // One thing to note before this check is that if num_bytes is greater than 64 (i.e
+            // exceeding RX FIFO's depth, then the below check will fail. For example, if
+            // wr_data_bytes contains 200 bytes and this function reads size_of(wr_data_bytes), then
+            // RX FIFO would contain the last 64 bytes that was read by the controller in rd_xfer().
+            if (wr_data_bytes[i] != rd_data_bytes[i]) {
```
engdoreis: this array is not needed, because you compare immediately after reading.
```suggestion
            if (wr_data_bytes[i] != i2c_rdata_byte(i2c)) {
```

sw/device/lib/hal/i2c.c
```diff
@@ -8,6 +8,14 @@
...
+// timing parameters.
+//
+// The values below are standard mode minimums
+volatile uint8_t SysClkPeriodNS = SYSCLK_NS;
+volatile uint16_t SCLLowTimeNs = 4700;
+volatile uint16_t HoldDataTimeNs = 1;
```
ziuziakowska: These should be `const` instead of `volatile`, and the names should be in `snake_case` with a prefix.

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "hal/i2c.h"
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+static bool wr_xfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_wr_bytes)
```
ziuziakowska: These should be `write_transfer`, `read_transfer`, etc.

sw/device/tests/i2c/host_test.c
ziuziakowska: These should be declared as `enum` constants at the top, e.g:
```c
enum : uint8_t {
    i2c_address = 0x3au,
    num_bytes = 0x8u,
};
```
the `host_tx_rx_test` should also use those values directly, especially for `num_bytes` to avoid creating a variable-length array.

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+
+bool test_main()
+{
+    i2c_t i2c = mocha_system_i2c();
+
+    // Configure the timing registers according to the I2C speed mode
+    i2c_init(i2c, i2c_speed_mode_standard);
+
+    // -- Configure IP for Controller mode --
+    i2c_enable_controller_mode(i2c);
```
ziuziakowska: The comments and extra lines here can be removed.

sw/device/tests/i2c/host_tx_rx_test.c
```diff
@@ -0,0 +1,80 @@
...
+    return (wr_xfer_status && rd_xfer_status);
+}
+
+static bool host_tx_rx_test(i2c_t i2c, uint8_t addr, uint8_t num_bytes)
+{
+    // Return status of the test
+    bool success = false;
+
+    // Data bytes to send to the target's receiver.
+    uint8_t wr_data_bytes[num_bytes];
```
ziuziakowska: It would be better to create a fixed-size array here, by using a constant in an `enum` declaration, instead of a value passed in by a function.

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,80 @@
...
+            // wr_data_bytes contains 200 bytes and this function reads size_of(wr_data_bytes), then
+            // RX FIFO would contain the last 64 bytes that was read by the controller in rd_xfer().
+            if (wr_data_bytes[i] != rd_data_bytes[i]) {
+                return false;
+            }
+        }
+        success = true;
+    }
+
+    return success;
```
ziuziakowska: The control flow here could be flattened a bit - in general it is better to check for *failure* conditions instead of success conditions, and return early if so, as that keeps the expected path/"happy path" easy to follow and without nesting, for example:
```suggestion
    if (!drive_xfer(i2c, addr, wr_data_bytes, num_bytes)) {
        return false;
    }

    for (uint8_t i = 0; i < num_bytes; i++) {
        if (wr_data_bytes[i] != i2c_rdata_byte(i2c)) {
            return false;
        }
    }

    return true;
```
This also removes the need for having and managing a `status` variable.

sw/device/tests/i2c/host_test.c
KinzaQamar: Yes, you are right as the `names` of the functions are loud enough to speak their functionality. Thanks

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
KinzaQamar: I think your point is valid;
> If this becomes the case then you are already ready for it and no change is required.

I'll add a rand keyword then

sw/device/lib/hal/i2c.c
KinzaQamar: I tried it with `SYSCLK_NS = 1ns`. So, the round-up divide will give us,

Standard mode:
SCL period = 10,000
rise_cycles = 450,
fall_cycles = 120,
low_cycles = 4000,
high_cycles = 4700(min) and 4730 after the above equation

Fast Mode:
SCL period = 2500
rise_cycles = 300, spec highest
fall_cycles = 300, spec highest
low_cycles = 1300,
high_cycles = 600(min) and satisfies the above equation

Fast Mode:
SCL period = 1000
rise_cycles = 120, spec highest
fall_cycles = 120, spec highest
low_cycles = 500,
high_cycles = 260(min) and satisfies the above equation

I believe if we violate OT's programming guide and chose greater `thigh` and `tlow` values, then the sum could overflow

KinzaQamar:
SW comments addressed

sw/device/lib/hal/i2c.c
elliotb-lowrisc: I now realise I wasn't thinking clearly when I wrote my previous comment. As you said earlier, the value of the period must be the sum of its parts (or close to it). Seems we'd have to be clocking at 65 GHz for that to overflow, so it seems save to leave it at 16 bits to me. Sorry for my confusion.

sw/device/tests/i2c/host_tx_rx_test.c
```diff
@@ -0,0 +1,67 @@
...
+    bool write_transfer_status = write_transfer(i2c, addr, data, num_bytes);
+    bool read_transfer_status = read_transfer(i2c, addr, num_bytes);
+
+    return (write_transfer_status && read_transfer_status);
+}
+
+static bool host_tx_rx_test(i2c_t i2c, uint8_t addr, uint8_t num_bytes)
+{
+    // Data bytes to send to the target's receiver.
+    uint8_t wr_data_bytes[num_bytes];
```
ziuziakowska: I realise now that this here seems a bit ambiguous - `num_bytes` could refer to either the function parameter or the `enum` constant 😅, I think it would be best to prefix the `enum` values with `test`, e.g `test_i2c_address`, `test_num_bytes`, and just using those values directly in the function by changing the signature to be `static bool host_tx_rx_test(i2c_t i2c)`

sw/device/tests/i2c/host_test.c
```diff
@@ -0,0 +1,67 @@
...
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+enum : uint8_t {
+    i2c_address = 0x3A,
+    num_bytes = 0x8,
+};
+
+static bool write_transfer(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_wr_bytes)
```
ziuziakowska: Like the other functions in `hal/i2c.h`, maybe this should be `num_bytes`?

sw/device/tests/i2c/host_tx_rx_test.c
```diff
@@ -8,6 +8,14 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+// Below symbols are going to be read by top_chip_dv_i2c_host_tx_rx_vseq in order to calculate agent
+// timing parameters.
+//
+// The values below are standard mode minimums
```
ziuziakowska:
```suggestion
// The below symbols are used by top_chip_dv_i2c_host_tx_rx_vseq in order to calculate agent timing parameters.
// The values are the minimums for standard mode.

**Review by ziuziakowska — COMMENTED** [review]
Just a few nits.

File: sw/device/tests/i2c/host_tx_rx_test.c
Comment (KinzaQamar) [reply sw/device/tests/i2c/host_tx_rx_test.c:1]: I renamed it. Does that look okay to you?

File: sw/device/tests/i2c/host_tx_rx_test.c (hunk 35 above)
Comment (KinzaQamar) [reply sw/device/tests/i2c/host_tx_rx_test.c]: Good spot, thanks!

File: sw/device/tests/i2c/host_tx_rx_test.c (hunk 37 above)
Comment (KinzaQamar) [reply sw/device/tests/i2c/host_tx_rx_test.c]: I've just rephrased it. Does that look okay to you?

File: sw/device/tests/i2c/host_tx_rx_test.c
Comment (KinzaQamar) [reply sw/device/tests/i2c/host_tx_rx_test.c:1]: Also, I've created #605 that renames existing i2c/smoketest.c

**Comment by KinzaQamar** [comment]
I've created #605 to address comments from @martin-velay and @elliotb-lowrisc on `hal/i2c.c`

File: sw/device/tests/i2c/host_tx_rx_test.c (hunk 37 above)
Comment (ziuziakowska) [reply sw/device/tests/i2c/host_tx_rx_test.c]: I think it still reads a bit off - the values aren't taken as spec minimums (that sounds like were deciding them), they are the spec minimums.

Comment (KinzaQamar) [reply sw/device/tests/i2c/host_tx_rx_test.c]: I've rephrased and shuffled the declarations

File: sw/device/tests/i2c/host_tx_rx_test.c [hunk 38]
```diff
@@ -0,0 +1,76 @@
...
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+// The const variables below are treated as symbols read by top_chip_dv_i2c_host_tx_rx_vseq in order
+// to calculate agent timing parameters.
+const uint8_t sys_clk_period_ns = SYSCLK_NS;
+
+// The constants are the spec minimums for standard mode speed expect hold_data_time_ns which should
```
elliotb-lowrisc: Possible typo? "expect" -> "except

elliotb-lowrisc (approved):
I've run the latest changes locally and found the new test passes with multiple seeds and the changes don't seem to break any other UVM top-level tests.

I did spot one typo, but it's not worth holding back this PR over

ziuziakowska (approved):

hw/top_chip/dv/top_chip_sim_cfg.hjson
martin-velay, line 415: I think the `_cheri` versions are missing here

hw/top_chip/dv/env/top_chip_dv_env_cfg.sv
```diff
@@ -45,6 +46,9 @@ function void top_chip_dv_env_cfg::initialize();
   // Configuration is required to perform meaningful monitoring
   m_uart_agent_cfg.en_tx_monitor = 0;
   m_uart_agent_cfg.en_rx_monitor = 0;
+
+  // Create I2C agent config object
+  m_i2c_agent_cfg = i2c_agent_cfg::type_id::create("m_i2c_agent_cfg");
```
martin-velay, line 51: I think it'd also be good to disable fcov collection at top level to avoid slowing down the simulations as this is not required as far as I know at this level:


```suggestion
  m_i2c_agent_cfg = i2c_agent_cfg::type_id::create("m_i2c_agent_cfg");
  // Functional coverage collection for chip level test is not needed for this agent
  m_i2c_agent_cfg.en_cov = 0;
```

hw/top_chip/dv/top_chip_sim_cfg.hjson
martin-velay, line 417: The `_cheri` versions are missing

hw/top_chip/dv/tb/tb.sv
```diff
@@ -260,6 +284,7 @@ module tb;
     uvm_config_db#(virtual clk_rst_if)::set(null, "*", "sys_clk_if", sys_clk_if);
     uvm_config_db#(virtual uart_if)::set(null, "*.env.m_uart_agent*", "vif", uart_if);
     uvm_config_db#(virtual pins_if #(NUM_GPIOS))::set(null, "*.env", "gpio_vif", gpio_pins_if);
+    uvm_config_db#(virtual i2c_if)::set(null, "*.env.m_i2c_agent", "vif", i2c_if);
```
martin-velay, line 287: Probably best to add a wildcard as if a sub-component of i2c agent tries to get this handle it will fail. But it doesn't hurt us to give this possibility in case:

```suggestion
    uvm_config_db#(virtual i2c_if)::set(null, "*.env.m_i2c_agent*", "vif", i2c_if);
```

martin-velay (commented):
A few more comments

hw/top_chip/dv/tb/tb.sv
KinzaQamar: `i2c_agent` gets this `vif` and assigned it to `cfg.vif`. But before that, this `cfg` is assigned to all the sub-components of `i2c_Agent` in `dv_base_agent`. Even if there will exist a new `sub-component` which requires that `vif`, it should follow the same convention for consistency.

hw/top_chip/dv/top_chip_sim_cfg.hjson
KinzaQamar: I'll add a `*_cheri` entry for `device` test in a top level I2C device PR

KinzaQamar: Good spot, thanks!

KinzaQamar:
Results: `dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i smoke -t xcelium`
```
### Test Results

|  Stage  |  Name  | Tests                               |  Max Job Runtime  |  Simulated Time  |  Passing  |  Total  |  Pass Rate  |
|:-------:|:------:|:------------------------------------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
|         |        | rv_timer_irq_cheri                  |     192.000s      |    2310.269us    |     1     |    1    |  100.00 %   |
|         |        | test_framework_exception_test_cheri |     163.000s      |    1591.614us    |     1     |    1    |  100.00 %   |
|         |        | gpio_smoke_cheri                    |     140.000s      |    1034.503us    |     1     |    1    |  100.00 %   |
|         |        | rv_timer_irq                        |     193.000s      |    2159.202us    |     1     |    1    |  100.00 %   |
|         |        | rv_plic_smoke_cheri                 |     137.000s      |    1075.209us    |     1     |    1    |  100.00 %   |
|         |        | gpio_smoke                          |     123.000s      |    886.820us     |     1     |    1    |  100.00 %   |
|         |        | uart_smoke_cheri                    |     135.000s      |    1678.177us    |     1     |    1    |  100.00 %   |
|         |        | clkmgr_smoke_cheri                  |     136.000s      |    1001.703us    |     1     |    1    |  100.00 %   |
|         |        | clkmgr_smoke                        |     124.000s      |    887.640us     |     1     |    1    |  100.00 %   |
|         |        | pwrmgr_smoke_cheri                  |     136.000s      |    1007.124us    |     1     |    1    |  100.00 %   |
|         |        | pwrmgr_smoke                        |     122.000s      |    884.710us     |     1     |    1    |  100.00 %   |
|         |        | mailbox_smoke                       |     122.000s      |    878.327us     |     1     |    1    |  100.00 %   |
|         |        | mailbox_smoke_cheri                 |     140.000s      |    1035.556us    |     1     |    1    |  100.00 %   |
|         |        | axi_sram_smoke                      |     120.000s      |    873.066us     |     1     |    1    |  100.00 %   |
|         |        | axi_sram_smoke_cheri                |     135.000s      |    989.682us     |     1     |    1    |  100.00 %   |
|         |        | i2c_host_tx_rx                      |     322.000s      |    2605.988us    |     1     |    1    |  100.00 %   |
|         |        | rstmgr_smoke                        |     223.000s      |    1679.884us    |     1     |    1    |  100.00 %   |
|         |        | rom_ctrl_smoke_cheri                |      1.293s       |     0.000us      |     0     |    1    |   0.00 %    |
|         |        | rv_timer_smoke_cheri                |     141.000s      |    1180.609us    |     1     |    1    |  100.00 %   |
|         |        | spi_device_smoke_cheri              |     156.000s      |    1269.145us    |     1     |    1    |  100.00 %   |
|         |        | rv_timer_smoke                      |     135.000s      |    1062.818us    |     1     |    1    |  100.00 %   |
|         |        | test_framework_exception_test       |     160.000s      |    1452.140us    |     1     |    1    |  100.00 %   |
|         |        | rom_ctrl_smoke                      |      1.249s       |     0.000us      |     0     |    1    |   0.00 %    |
|         |        | rv_plic_smoke                       |     117.000s      |    910.759us     |     1     |    1    |  100.00 %   |
|         |        | spi_host_smoke_cheri                |      1.362s       |     0.000us      |     0     |    1    |   0.00 %    |
|         |        | rstmgr_smoke_cheri                  |     251.000s      |    1906.446us    |     1     |    1    |  100.00 %   |
|         |        | spi_host_smoke                      |      1.430s       |     0.000us      |     0     |    1    |   0.00 %    |
|         |        | entropy_src_smoke                   |     162.000s      |    1169.361us    |     1     |    1    |  100.00 %   |
|         |        | i2c_host_tx_rx_cheri                |     328.000s      |    2728.893us    |     1     |    1    |  100.00 %   |
|         |        | spi_device_smoke                    |     128.000s      |    1067.820us    |     1     |    1    |  100.00 %   |
|         |        | uart_smoke                          |     150.000s      |    1521.959us    |     1     |    1    |  100.00 %   |
|         |        | entropy_src_smoke_cheri             |     182.000s      |    1317.569us    |     1     |    1    |  100.00 %   |
|         |        | **TOTAL**                           |                   |                  |    28     |   32    |   87.50 %   |
|         |        | **TOTAL**                           |                   |                  |    28     |   32    |   87.50 %   |
```

martin-velay (approved):
It looks good to me, thanks Kinza

PR #474, opened 2026-04-27
[hw] DRAM tag base and mask fixed

### LINKED ISSUE #471: Out of bounds accesses not causing access faults
Reading/writing outside the range of memory does not cause access faults to be raised.

For a simple reproducer, the tag controller smoketest can be extended like so:
```diff
diff --git a/sw/device/tests/tag_controller/smoketest.c b/sw/device/tests/tag_controller/smoketest.c
index c2e12f2..c078c83 100644
--- a/sw/device/tests/tag_controller/smoketest.c
+++ b/sw/device/tests/tag_controller/smoketest.c
@@ -4,6 +4,8 @@
 
 #include "hal/mmio.h"
 #include "hal/mocha.h"
+#include "hal/uart.h"
+#include "runtime/print.h"
 #include <stdbool.h>
 #include <stdint.h>
 
@@ -26,6 +28,12 @@ bool data_readback_test(void *dram)
         return false;
     }
 
+    uart_t console = mocha_system_uart();
+    uart_init(console);
+    // just out of range...
+    uint64_t v = DEV_READ64(dram + 0x3F800000);
+    uprintf(console, "v = %lx\n", v);
+
     // Varying sizes
     DEV_WRITE8(dram + 5, 0x5A);
     if (DEV_READ8(dram + 5) != 0x5A) {
```

When the Vanilla test is ran in Verilator, the following gets printed to the UART, showing the read was successful:
```
v = ca11ab1ebadcab1e
TEST RESULT: PASSED
```
It is expected that these raise a Load/Store Access Fault (and Instruction Access Fault for instruction fetch).
Note that this would also affect CHERI targets if any capability bounds include memory that is out-of-range.

marnovandermaas:
Previously we had conflated the ideas of the actual physical length of DRAM and the DRAM length minus the tag store. This caused multiple issues in the design.

Closes: https://github.com/lowRISC/mocha/issues/471

### DIFF
--- a/hw/top_chip/rtl/top_chip_system.sv
+++ b/hw/top_chip/rtl/top_chip_system.sv
@@ -116,7 +116,7 @@ module top_chip_system #(
     '{ idx: top_pkg::SRAM,       start_addr: top_pkg::SRAMBase,       end_addr: top_pkg::SRAMBase       + top_pkg::SRAMLength       },
     '{ idx: top_pkg::Mailbox,    start_addr: top_pkg::MailboxBase,    end_addr: top_pkg::MailboxBase    + top_pkg::MailboxLength    },
     '{ idx: top_pkg::TlCrossbar, start_addr: top_pkg::TlCrossbarBase, end_addr: top_pkg::TlCrossbarBase + top_pkg::TlCrossbarLength },
-    '{ idx: top_pkg::DRAM,       start_addr: top_pkg::DRAMBase,       end_addr: top_pkg::DRAMBase       + top_pkg::DRAMLength       }
+    '{ idx: top_pkg::DRAM,       start_addr: top_pkg::DRAMBase,       end_addr: top_pkg::DRAMBase       + top_pkg::DRAMUsableLength }
   };
 
   // TileLink signals.
@@ -1011,7 +1011,7 @@ module top_chip_system #(
   // Instantiate CHERI tag controller for DRAM
   axi_tagctrl_reg_wrap #(
     .DRAMMemBase      ( 32'(top_pkg::DRAMBase)            ),
-    .DRAMMemLength    ( 32'(top_pkg::DRAMLength)          ),
+    .DRAMMemLength    ( 32'(top_pkg::DRAMPhysicalLength)  ),
     .CapSize          ( top_pkg::CapSizeBits              ),
     .TagCacheMemBase  ( 32'(top_pkg::TagCacheMemBase)     ),
     .SetAssociativity ( top_pkg::TagCacheSetAssociativity ),
--- a/hw/top_chip/rtl/top_chip_system.sv
+++ b/hw/top_chip/rtl/top_chip_system.sv
@@ -76,11 +76,21 @@ module top_chip_system #(
   // CVA6 configuration
   function automatic config_pkg::cva6_cfg_t build_cva6_config(config_pkg::cva6_user_cfg_t CVA6UserCfg);
     config_pkg::cva6_user_cfg_t cfg = CVA6UserCfg;
-    cfg.RVZiCond              = bit'(0);
-    cfg.CvxifEn               = bit'(0);
-    cfg.NrNonIdempotentRules  = unsigned'(1);
-    cfg.NonIdempotentAddrBase = 1024'({64'b0});
-    cfg.NonIdempotentLength   = 1024'({top_pkg::SRAMBase});
+    cfg.RVZiCond                    = bit'(0);
+    cfg.CvxifEn                     = bit'(0);
+    cfg.DmBaseAddress               = top_pkg::DebugMemBase;
+    cfg.NrExecuteRegionRules        = unsigned'(4);
+    cfg.ExecuteRegionAddrBase       = 1024'({top_pkg::DRAMBase,
+                                             top_pkg::DebugMemBase,
+                                             top_pkg::SRAMBase,
+                                             top_pkg::RomCtrlMemBase});
+    cfg.ExecuteRegionLength         = 1024'({top_pkg::DRAMUsableLength,
+                                             top_pkg::DebugMemLength,
+                                             top_pkg::SRAMLength,
+                                             top_pkg::RomCtrlMemLength});
+    cfg.NrCachedRegionRules         = unsigned'(1);
+    cfg.CachedRegionAddrBase        = 1024'({top_pkg::DRAMBase});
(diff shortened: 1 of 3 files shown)


ziuziakowska:
This seems correct now, with `0x40000000` being the "real" length of DRAM, and `0x3f800000` being the "usable" portion:
```
> 0x40000000 >> log(128, 2) to hex

  shift(1073741824, −log(128, 2)) = 0x800000

> 0x40000000 - 0x800000 to hex

  1073741824 − 8388608 = 0x3F800000
```

hw/top_chip/rtl/top_pkg.sv
```diff
@@ -46,18 +46,19 @@ package top_pkg;
   localparam longint unsigned SRAMLength        = 64'h0002_0000;
   localparam longint unsigned MailboxLength     = 64'h0001_0000;
   localparam longint unsigned TlCrossbarLength  = 64'h1000_0000;
-  localparam longint unsigned DRAMLength        = 64'h3F80_0000;
+  localparam longint unsigned DRAMActualLength  = 64'h4000_0000;
```
raylau1: It may be more clear to rename `DRAMActualLength` to `DRAMPhysicalLength`

hw/top_chip/rtl/top_pkg.sv
```diff
@@ -46,18 +46,19 @@ package top_pkg;
...
   localparam longint unsigned TlCrossbarMask    = TlCrossbarLength - 1;
-  localparam longint unsigned DRAMMask          = DRAMLength - 1;
+  localparam longint unsigned DRAMMask          = DRAMActualLength - 1;
 
   // Tag controller parameters
   localparam int     unsigned CapSizeBits              = 128;
-  localparam longint unsigned TagCacheMemLength        = DRAMLength >> $clog2(CapSizeBits);
-  localparam longint unsigned TagCacheMemBase          = DRAMBase + DRAMLength - TagCacheMemLength;
+  localparam longint unsigned TagCacheMemLength        = DRAMActualLength >> $clog2(CapSizeBits);
+  localparam longint unsigned DRAMLength               = DRAMActualLength - TagCacheMemLength;
```
raylau1: It may be more clear to rename `DRAMLength` to `DRAMUsableLength`

raylau1 (commented):
A few naming nits. Also we should keep the CVA6 config up-to-date with the named address ranges, perhaps by adding the following to the CVA6 config function in `top_chip_system`:
```systemverilog
    cfg.NrExecuteRegionRules  = unsigned'(2);
    cfg.ExecuteRegionAddrBase = 1024'({top_pkg::DRAMBase, top_pkg::SRAMBase});
    cfg.ExecuteRegionLength   = 1024'({top_pkg::DRAMUsableLength, top_pkg::SRAMLength});
    cfg.NrCachedRegionRules   = unsigned'(1);
    cfg.CachedRegionAddrBase  = 1024'({top_pkg::DRAMBase});
    cfg.CachedRegionLength    = 1024'({top_pkg::DRAMUsableLength});
```

ziuziakowska (commented):
I believe this also needs updating in the CVA6 config (it should use the usable portion, not the full length with the tag cache), this might fix https://github.com/lowRISC/mocha/issues/471.

Maybe it would also be useful to add some assertions based on the lengths, like:
- Usable DRAM length + Tag Cache length = Total DRAM length
- [DRAM base, DRAM base + Usable DRAM length) and [Tag cache base, Tag cache base + Tag cache length) are non-overlapping regions

marnovandermaas:
Related to this I've created a PR to change the executable regions and cached regions accordingly: https://github.com/lowRISC/cva6-cheri/pull/2/changes

hw/top_chip/rtl/top_pkg.sv
marnovandermaas: Good point. I'll make this change.

hw/top_chip/rtl/top_pkg.sv
marnovandermaas: Good point, I'll make this change.

marnovandermaas:
Marking as draft because bitstream build fails.

raylau1:
Continuing with the split top package may cause a lot of code churn. I think it'll be simpler if we left the upstream CVA6 config at default (with unnamed constants) and override the config in `build_cva6_config` in `top_chip_system`. This should avoid the circular dependency and code churn from `top_bus_pkg`.

hw/top_chip/rtl/top_chip_system.sv
```diff
@@ -76,11 +76,24 @@ module top_chip_system #(
...
-    cfg.RVZiCond              = bit'(0);
-    cfg.CvxifEn               = bit'(0);
-    cfg.NrNonIdempotentRules  = unsigned'(1);
-    cfg.NonIdempotentAddrBase = 1024'({64'b0});
-    cfg.NonIdempotentLength   = 1024'({top_pkg::SRAMBase});
+    cfg.RVZiCond                    = bit'(0);
+    cfg.CvxifEn                     = bit'(0);
+    cfg.NrNonIdempotentRules        = unsigned'(1);
+    cfg.NonIdempotentAddrBase       = 1024'({64'b0});
+    cfg.NonIdempotentLength         = 1024'({top_pkg::RomCtrlMemBase});
```
raylau1: If we are not specifying non-idempotent address ranges, the length should be 0 like the default config

marnovandermaas: Yes, I'll probably just remove these overrides.

marnovandermaas:
Just double checked this doesn't trivially break top-level DV:
```
### Branch: dram_fix
### [Testplan](hw/top_chip/data/top_mocha_sim_testplan.hjson)
### Simulator: XCELIUM

### Test Results

|  Stage  |  Name  | Tests      |  Max Job Runtime  |  Simulated Time  |  Passing  |  Total  |  Pass Rate  |
|:-------:|:------:|:-----------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
|         |        | uart_smoke |      10.000s      |     0.000us      |     1     |    1    |  100.00 %   |
|         |        | **TOTAL**  |                   |                  |     1     |    1    |  100.00 %   |
|         |        | **TOTAL**  |                   |                  |     1     |    1    |  100.00 %   |
```

raylau1 (approved):
Thanks for the changes. LGTM!

PR #475, opened 2026-04-28
[sram] Writes accompanied by defined read data

marnovandermaas:
The axi_to_detailed_mem module requires a valid read request to be accompanied by every write. This is problematic since at the start of time the content of SRAM is understandably undefined. Instead of feeding don't cares through the module alongside an rvalid signal. Now this feeds a recognisable hex sequence through the module so that it still works but it is also somewhat detectable if this read propagates to the output.

This change is mainly to unblock: https://github.com/lowRISC/mocha/pull/447


marnovandermaas:
Closing this as it was already merged in the boot ROM PR.

PR #504, opened 2026-04-29
FPGA Runner: Remove timeout handling, add `run` subcommand

ziuziakowska:
This PR makes some changes to the FPGA runner.

Test timeout handling is removed as this functionality is now handled by CMake/CTest, so the runner can be simplified to not do this itself.

This PR also moves the testing aspect of the runner to a `test` subcommand, and introduces a `run` subcommand which just loads a program over SPI and exits. This is useful for simply loading and running interactive software, such as examples and demos, and in the future interactive Operating Systems. This subcommand does not attempt to open the UART, which means that the UART can be kept open continuously across reboots in a serial console e.g screen or picocom, which is quite convenient.

### DIFF
--- a/util/fpga_runner.py
+++ b/util/fpga_runner.py
@@ -42,14 +42,14 @@ async def set_pin(pin: int, val: bool) -> None:
             print(f"[{RUNNER}] gpio-write command exited with non-zero exit code {p.returncode}")
             sys.exit(1)
     except TimeoutError:
-        print(f"[{RUNNER}] gpio-write timed out after {FTDITOOL_WAIT_TIME} seconds.")
+        print(f"[{RUNNER}] gpio-write timed out after {FTDITOOL_WAIT_TIME} seconds")
         await p.terminate()  # or maybe p.kill()
         sys.exit(1)
 
 
 async def reset_core() -> None:
     """
-    Pull the reset pin down for 100ms.
+    Reset the core by pulling the reset pin down for 100ms.
     """
     await set_pin(2, False)
     await asyncio.sleep(0.1)
@@ -79,13 +79,11 @@ def get_load_addr(elf: Path) -> int:
                 seg["p_paddr"] for seg in elf.iter_segments() if seg["p_type"] == "PT_LOAD"
             ]
             if not load_addrs:
-                print(f"[{RUNNER}] no `PT_LOAD` segments found in elf {elf}")
+                print(f"[{RUNNER}] no PT_LOAD segments found in ELF file {elf}")
                 sys.exit(1)
-
-            first_address = min(load_addrs)
-            return first_address
+            return min(load_addrs)
     except OSError as e:
-        print(f"[{RUNNER}] error opening elf {elf}: {e}")
+        print(f"[{RUNNER}] error opening ELF file {elf}: {e}")
         sys.exit(1)
 
 
--- a/util/fpga_runner.py
+++ b/util/fpga_runner.py
@@ -22,17 +22,18 @@ MOCHA_BOOTROM_BOOSTRAP_STR = "Entering SPI bootstrap"
 RUNNER: str = Path(__file__).name
 
 
-async def set_pin(pin: int, val: int) -> None:
+async def set_pin(pin: int, val: bool) -> None:
     command = [
         "ftditool",
         "--pid",
         hex(FTDI_PID),
         "gpio-write",
         str(pin),
-        str(val),
+        str(int(val)),
         "--ftdi",
         FTDI_DEVICE_DESC,
     ]
+
     p = await asyncio.create_subprocess_exec(*command)
     try:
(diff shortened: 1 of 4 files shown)


util/fpga_runner.py
```diff
@@ -55,14 +56,18 @@ async def reset_core() -> None:
...
-    Pull down the bootstrap pin, reset the core, wait for the bootstrap string over uart,
-    then release the pin the bootstrap pin.
+    Pull down the bootstrap pin and reset the core. If a UART is provided,
+    wait for the bootstrap string over UART, otherwise wait a short period
+    of time and assume the system is ready for bootstrapping. Then release the pin.
     """
     await set_pin(0, False)
     await reset_core()
-    await poll_uart_checking_for(uart, MOCHA_BOOTROM_BOOSTRAP_STR)
+    if uart is not None:
```
engdoreis, line 67: ~~Why not use the uart for the run subcommand?~~
I just read the PR header, I think that there should be an arg to not use uart for the run subcommand `--no-uart`, or try to open it and if fails use the delay.

ziuziakowska: I think that that would add a bit more complexity, and for the use-case this is intended for, not using the UART would almost always be preferred (indeed locally I found myself always commenting out the code that checked for the ROM bootstrapping and test status strings).

util/fpga_runner.py
```diff
@@ -84,35 +89,30 @@ def get_load_addr(elf: Path) -> int:
...
         print(f"[{RUNNER}] SPI load command exited with non-zero exit code {p.returncode}")
         sys.exit(1)
 
 
-async def run_fpga_test(tty: str, test: Path) -> bool:
-    with serial.Serial(tty, BAUD_RATE, timeout=0) as uart:
-        await load_fpga_test(test, uart)
-        pattern = r"TEST RESULT: (PASSED|FAILED)"
-        result = await asyncio.create_task(poll_uart_checking_for(uart, pattern))
-        return "PASSED" in result
```
marnovandermaas: Nit: you're changing this line twice in the same PR. Is there any way to avoid this?

util/fpga_runner.py
```diff
@@ -130,21 +130,63 @@ def find_uart(vid: int = 0x0403, pid: int = 0x6001) -> str | None:
...
+    )
+
+    run_parser = subparsers.add_parser("run", help=run_help, description=("run: " + run_help))
+    run_parser.add_argument("path", type=Path, help="Path to program ELF to run on the FPGA")
+    run_parser.set_defaults(func=do_fpga_run)
+
+    args = parser.parse_args()
+    try:
+        asyncio.run(args.func(args))
+    except KeyboardInterrupt:
```
marnovandermaas: Is this not the default behaviour of Python?

marnovandermaas (commented):
Nice PR. Just a few minor comments from me (non blocking in case Douglas approves)

util/fpga_runner.py
engdoreis: I think that we can recommend this script in the release documentation to load the examples, so it would be nice to print the uart output by default. So, for development, one just need to open the uart prior to running this script.

engdoreis (approved):
LGTM.

util/fpga_runner.py
```diff
@@ -136,21 +129,63 @@ def find_uart(vid: int = 0x0403, pid: int = 0x6001) -> str | None:
...
+    """
     if uart_tty := find_uart():
-        try:
-            success = asyncio.run(run_fpga_test(uart_tty, args.test))
-            sys.exit(0 if success else 1)
-        except KeyboardInterrupt:
+        with serial.Serial(uart_tty, BAUD_RATE, timeout=0) as uart:
+            await load_fpga_binary(args.path, uart)
+
+            result = await asyncio.create_task(poll_uart_checking_for(uart, TEST_PATTERN))
```
AlexJones0: I probably missed this in the previous PR that introduced the code being moved here, but do you know if there is a reason why we `asyncio.create_task` and immediately `await` it? Isn't this redundant here, i.e.:

```suggestion
            result = await poll_uart_checking_for(uart, TEST_PATTERN)
```

util/fpga_runner.py
AlexJones0: I'm also not sure I fully understand - is this just to suppress the error traceback when you e.g. Ctrl-C?

AlexJones0 (approved):

util/fpga_runner.py
ziuziakowska: I think this is redundant, thanks for catching that

util/fpga_runner.py
ziuziakowska: Yes, I'll add a comment about that :+1:

AlexJones0 (approved):
Thanks @ziuziakowska, this LGTM.

PR #505, opened 2026-04-29
[fpga] Remove duplicate RomInitFile parameter

raylau1:
This removes the duplicate `RomInitFile` parameter in the chip top core file.


marnovandermaas (approved):
Good catch.

PR #512, opened 2026-04-30
[fpga,sw] Add ethernet MAC to Genesys 2 top

### LINKED ISSUE #354: Integrate ethernet to Genesys 2 top

raylau1:
This PR integrates the ariane-ethernet MAC core on a rest of chip AXI crossbar outside Mocha. ~RDL,~ HAL driver, smoketest and example software are also added.

Closes #354 

### Testing the ethernet MAC
Launch wireshark and begin capturing packets on the wired interface, make sure wireshark has sufficient permissions to capture packets.

Once the bitstream is loaded, run the ethernet example:
```sh
./util/fpga_runner.py build/sw/device/examples/ethernet_example
```

You should see the following UART output:
```
Ethernet example
Sending packets
Receiving packets
```

Your should also see 10 packets captured in wireshark with the following sizes: `64, 1500, 1500, 1500, 100, 64, 1500, 1500, 1500, 100`.

To test packet reception, launch python as **root**, which is necessary to send raw packets.

Then bind to the wired interface with:
```python3
from socket import *
s = socket(AF_PACKET, SOCK_RAW)
s.bind(("<wired interface>", 0))
```

You should now be able to send packets to Genesys 2. For example, running the following:
```python3
s.send(b"\x9A\xBC\x12\x34\x56\x78" + b"\xFF"*(1500-6))
```
Sends a 1500 byte packet to Genesys 2 with a destination MAC address of `9A:BC:12:34:56:78`.

You should see the packet length and content reported on the UART output, note that the length and content include the 4-byte FCS.


hw/top_chip/rtl/top_pkg.sv
```diff
@@ -48,16 +50,41 @@ package top_pkg;
...
+  typedef enum int unsigned {
+    MochaAXICrossbar = 0
+  } rest_of_chip_axi_hosts_t;
+
+  typedef enum int unsigned {
+    Ethernet = 0
+  } rest_of_chip_axi_devices_t;
+
+  typedef enum longint unsigned {
+    EthernetBase = 64'h3000_0000
```
marnovandermaas, line 79: This should be added to Mocha RDL as well so it appears in the memory map.

hw/top_chip/data/pins_genesys2.xdc
```diff
@@ -51,6 +51,23 @@ set_property -dict { PACKAGE_PIN T27   IOSTANDARD LVCMOS33 PULLTYPE PULLUP } [ge
...
+## Ethernet RGMII
+set_property -dict { PACKAGE_PIN AH24  IOSTANDARD LVCMOS33 } [get_ports { eth_rgmii_rst_no   }];
+set_property -dict { PACKAGE_PIN AG10  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxck_i   }];
+set_property -dict { PACKAGE_PIN AH11  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxctl_i  }];
+set_property -dict { PACKAGE_PIN AJ14  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxd_i[0] }];
+set_property -dict { PACKAGE_PIN AH14  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxd_i[1] }];
+set_property -dict { PACKAGE_PIN AK13  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxd_i[2] }];
+set_property -dict { PACKAGE_PIN AJ13  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_rxd_i[3] }];
+set_property -dict { PACKAGE_PIN AE10  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_txck_o   }];
+set_property -dict { PACKAGE_PIN AK14  IOSTANDARD LVCMOS15 } [get_ports { eth_rgmii_txctl_o  }];
```
marnovandermaas: This one is marked as ETH_TX_EN in the datasheet, but I don't see an ETH_TX_CTL so I guess this is ok.

hw/top_chip/rtl/ethernet_wrapper.sv
```diff
@@ -0,0 +1,156 @@
...
+    // Memory interface.
+    .mem_req_o    (eth_en),
+    .mem_gnt_i    (1'b1),
+    .mem_addr_o   (eth_addr),
+    .mem_wdata_o  (eth_wrdata),
+    .mem_strb_o   (eth_be),
+    .mem_atop_o   ( ),
+    .mem_we_o     (eth_we),
+    .mem_rvalid_i (eth_rvalid),
+    .mem_rdata_i  (eth_rdata)
```
marnovandermaas: Should we do something like the following to mark bad read datas?
`(eth_rvalid & eth_we) ? 64'hBADDBADDBADDBADD : eth_rdata`

hw/top_chip/rtl/ethernet_wrapper.sv
```diff
@@ -0,0 +1,156 @@
...
+  );
+
+  // Single-cycle read response.
+  always_ff @(posedge clk_125m_i or negedge rst_eth_ni) begin
+    if (!rst_eth_ni) begin
+      eth_rvalid <= '0;
+    end else begin
+      eth_rvalid <= eth_en; // Generate rvalid strobes even for writes
```
marnovandermaas, line 109: I assume this is necessary for the axi_to_mem like it is in our SRAM.

hw/top_chip/rtl/ethernet_wrapper.sv
```diff
@@ -0,0 +1,156 @@
...
+  output logic       eth_rgmii_mdc_o
+);
+  // AXI signals from CDC FIFO to axi_to_mem, synchronous to clk_125m
+  top_pkg::axi_req_t  eth_125m_req;
+  top_pkg::axi_resp_t eth_125m_resp;
+
+  // Framing top 64-bit memory format signals
+  logic                               eth_en;
+  logic [  top_pkg::AxiAddrWidth-1:0] eth_addr;
+  logic [  top_pkg::AxiDataWidth-1:0] eth_wrdata;
```
marnovandermaas: Nit: can we make this wdata to match rdata?

hw/vendor/ethernet/eth_mac_1g_rgmii.sv
```diff
@@ -130,8 +130,12 @@ end
 // PHY speed detection
 reg [2:0] rx_prescale;
 
-always @(posedge rx_clk) begin
-    rx_prescale <= rx_prescale + 3'd1;
+always @(posedge rx_clk or posedge gtx_rst) begin
```
marnovandermaas: Are we sure that rx_clk is synchronised to gtx_clk since we are using gtx_rst here?

hw/vendor/patches/ethernet/0007_Fix_Rx_Packet_First_8_Bytes_Incorrect.patch
marnovandermaas, line 1: Just from my understanding. If there is a packet that is not accepted, the first 8 bytes of the next RX buffer are written. Then based on that the IP decides whether to drop the packet. Does the nextbuf then stay the same because the packet was rejected? In which case if an accepted packed follows this rejected packet the first 8 bytes will be overwritten again by the accepted packet? Or does the nextbuf get incremented anyway and then the read buffer has this rejected packet inside it?

hw/vendor/ethernet/framing_top.sv
```diff
@@ -116,11 +116,20 @@ reg        byte_sync, sync, irq_en, tx_busy;
    assign tx_axis_tdata = douta >> {tx_frame_addr[2],3'b000};
    assign phy_mdc = phy_mdclk;
    
+    logic       rx_use_incr_buf_offset;
+    logic [2:0] rx_buf_offset;
+
+    always_ff @(posedge clk_int) begin
+      rx_use_incr_buf_offset <= (rx_addr_axis < 7);
```
marnovandermaas: Does this need to check the value of rst_int?

sw/device/lib/hal/ethernet.h
marnovandermaas, line 1: How come you added this as well as the auto-generated version? The RDL version is preferable but we can do that in a later PR. I do feel relatively strongly about that we shouldn't have both the RDL as well as the manually written HAL.

marnovandermaas (commented):
This is amazing work. I'll need to test this before I can approve it. In the meantime I've left some minor comments.

hw/top_chip/rtl/top_pkg.sv
marnovandermaas: This was done in a later commit.

hw/top_chip/rtl/ethernet_wrapper.sv
raylau1: Yes

hw/top_chip/rtl/ethernet_wrapper.sv
raylau1: Yes I'll add something similar but I think it's not necessary to include `eth_rvalid` in the condition

hw/vendor/patches/ethernet/0007_Fix_Rx_Packet_First_8_Bytes_Incorrect.patch
raylau1: The basic idea is that received packet data is always written into next buffer, and the nextbuf pointer is only incremented if this packet is accepted. This means if an accepted packet follows a rejected packet, it will overwrite the rejected packet data in next buffer and only then is the nextbuf pointer incremented.

In reality, the framing top increments nextbuf immediately after the MAC address check for an accepted packet, and with this patch the first 8 bytes of a received packet is always preemptively written into the buffer at (nextbuf+1), then nextbuf is possibly incremented, then the rest of the packet is written into the buffer at the new nextbuf.

This means an accepted packet is fully written at new nextbuf (i.e. old nextbuf + 1), while for a dropped packet, the first 8 bytes is written at (old nextbuf + 1), and the rest is written at old nextbuf, with nextbuf not changing after the fact.

To compensate for this, the software driver reads from (firstbuf+1), then increments firstbuf.

hw/vendor/ethernet/eth_mac_1g_rgmii.sv
raylau1: It isn't, I'll add a synchroniser for the reset here

hw/vendor/ethernet/framing_top.sv
raylau1: Hmm I think it's better to add a reset here, I'll add one.

sw/device/lib/hal/ethernet.h
raylau1: I didn't use the auto generated header for the driver because I couldn't use a buffer memory width of 64 bits and wasn't sure if the framing top can work with 32-bit accesses. I am happy to spend some time and try moving to the RDL version though.

hw/top_chip/data/pins_genesys2.xdc
raylau1: I do think it'll be better if the port names are more consistent with the schematic, so I'll be fixing this.

hw/top_chip/rtl/ethernet_wrapper.sv
raylau1: Yes, I'll add a fix

hw/top_chip/rtl/ethernet_wrapper.sv
marnovandermaas: Great, I'll mark this as resolved.

hw/vendor/patches/ethernet/0007_Fix_Rx_Packet_First_8_Bytes_Incorrect.patch
marnovandermaas: It's kind of sad that the rest of the data for a dropped packet overwrites the rest of "old nextbuf". This is essentially overwriting the body from the previous valid packet right? Does this not cause any issues when rejected and accepted packets are interleaved?

sw/device/lib/hal/ethernet.h
marnovandermaas: I'm happy to keep your manually written file here to move this PR forward. Then we can move to the RDL later once the 64-bit accesses has been fixed.

hw/vendor/patches/ethernet/0007_Fix_Rx_Packet_First_8_Bytes_Incorrect.patch
raylau1: Yes it will overwrite the previous valid packet if there is at least 1 valid packet in the buffers. This has not happened in my testing since the packet bandwidth is too low for a packet to arrive before the previous has been completely processed. One way to work around this issue is to enable promiscuous mode and do MAC filtering in the driver.

raylau1: I have just reproduced this with interleaved 1500-byte packets, so a workaround is necessary at high bandwidths

hw/top_chip/data/pins_genesys2.xdc
raylau1: This is now fixed

hw/top_chip/rtl/ethernet_wrapper.sv
raylau1: This is now fixed

hw/vendor/ethernet/eth_mac_1g_rgmii.sv
raylau1: This is now fixed

hw/vendor/ethernet/framing_top.sv
raylau1: This is now fixed

sw/device/lib/hal/ethernet.h
raylau1: Ethernet RDL is now moved to the [ethernet_rdl](https://github.com/lowRISC/mocha/tree/ethernet_rdl) branch. Issue #522 is open regarding the 64-bit memwidth.

sw/device/examples/ethernet_example.c
```diff
@@ -0,0 +1,114 @@
...
+        // Increment first buffer ID (4 bits)
+        firstbuf = (firstbuf + 1) & 0b1111;
+
+        // Get packet length
+        // This includes an additional 4 bytes for the FCS
+        len = ethernet_rx_buffer_packet_length_get(ethernet, firstbuf & 0b111);
+        len_words = len / sizeof(uint64_t);
+        if ((len % sizeof(uint64_t)) != 0) {
+            ++len_words;
+        }
```
engdoreis: Nit: This will also round up.
```suggestion
        len_words =  (len + sizeof(uint64_t) - 1) / sizeof(uint64_t);
```

raylau1: Thanks, this is now fixed

engdoreis:
I tested and I can see packages being sent and recived.

engdoreis (approved):
LGTM.

sw/device/lib/hal/ethernet.h
marnovandermaas: Thank you.

hw/vendor/patches/ethernet/0007_Fix_Rx_Packet_First_8_Bytes_Incorrect.patch
marnovandermaas: Ok, I think this is fine for this initial pull request. Shall we open an issue to track this for high bandwidth?

raylau1: On second thought I can create a better patch that doesn't have this issue in a few hours

marnovandermaas (approved):
I'm happy for this to go in as I did a detailed code review before and Douglas has tested on FPGA.

PR #520, opened 2026-05-04
[hw] CVA6 updated with HPDCache

### LINKED ISSUE #245: CVA6-CHERI with HPDCache
This is related to #62

The first case of doing this is to integrate CVA6-CHERI snapshot that uses HPDCache as data cache. Currently the cache sub-system is hacked to work only when the data-cache is disabled. This issue is captured upstream here: https://github.com/Capabilities-Limited/cheri-cva6/issues/10

marnovandermaas:
This is the first major update of CVA6 since our initial integration. The core has seen quite a few fixes and most importantly the HPDCache is now integrated for use.

Closes: https://github.com/lowRISC/mocha/issues/245


hw/vendor/cva6_cheri/core/cache_subsystem/wt_cache_subsystem.sv
tchilikov-semify, line 1: If we intend to merge the testrig harness into mocha, its worth noting that the rvfi_dii_generator module within this generate block is currently used by the harness for instruction injection. In this case since the ports are just a subset of the icache ports, I think it should be possible to hook the injector into the icache fairly easily. 

The only challenge in this case is omitting the real icache during elaboration, and taking the DII instruction injector instead, which could be solved via a systemverilog config file (if we dont want to change the .core files). This would add a little bit of complexity, but could still work cleanly.

Alternatively, we could leave the generator instantiation in.

Both approaches are OK from my perspective, just a matter of taste and weather we are OK with having simulation-only modules instantiated explicitly inside the RTL.

hw/top_chip/rtl/top_chip_system.sv
tchilikov-semify, line 1: I noticed we have the config overrides in the top level of mocha before as well, just wondering why do we do this instead of changing the config directly?

hw/top_chip/lint/top_chip_system.vlt
```diff
@@ -6,8 +6,12 @@
 
 `verilator_config
 
+// Turn off UNOPTFLAT warnings in the whole design.
+// We may turn these back on once the design is more finalised but for now they give too many false positives.
+lint_off -rule UNOPTFLAT
+
```
raylau1: Can we turn off this warning globally for only the CVA6 and HPDCache? I prefer to have this check enabled for top-related RTL.

raylau1 (commented):
The hardware changes look good, though there are DV failures about FIFO unknown data that needs to be resolved first.

Nit: The link in commit 9e623b70da2b2b2e7eb24b69c5e5e617b47b9bcd description is broken because it is split into 2 lines

hw/top_chip/rtl/top_chip_system.sv
raylau1: This is so that we can use named constants from top_pkg. If we were to modify the config directly we will either have to use unnamed constants or introduce a circular dependency which is difficult to resolve without a lot of extra complexity. This is from the [discussions in PR 474](https://github.com/lowRISC/mocha/pull/474#issuecomment-4327668276) when we tried to do this.

tchilikov-semify: clear now - thanks for the explanation Ray!

hw/top_chip/lint/top_chip_system.vlt
marnovandermaas: Agreed, I've just put the specific UNOPTFLAT warnings in for HPDCache. They weren't that many.

hw/vendor/cva6_cheri/core/cache_subsystem/wt_cache_subsystem.sv
marnovandermaas: I think for now we can leave it in but I agree that in the long run we should not have simulation only modules corrupting our RTL.

marnovandermaas:
I'm marking this as draft because there are X propagation issues in CHERI mode.

marnovandermaas:
I've fixed the X-propagation issue and opened a pull request upstream: https://github.com/Capabilities-Limited/cheri-cva6/pull/79
I'm still keeping this as draft since for some reason there is a timeout in the DV environment. It looks like there is an infinite loop of exceptions at the end causing the timeout.

marnovandermaas:
Ok, this is now passing the regression smoke test with only the same failures as on main: SPI device and ROM control.

hw/top_chip/rtl/axi_sram.sv
```diff
@@ -84,17 +84,17 @@ module axi_sram #(
...
 
   // Tag bit address calculation
-  assign sram_tag_bit_addr   = TagBitAddrWidth'((sram_addr & top_pkg::SRAMMask) >>
+  assign sram_tag_bit_addr     = TagBitAddrWidth'((sram_addr & top_pkg::SRAMMask) >>
                                $clog2(top_pkg::CapSizeBits / 8));
-  assign sram_tag_word_addr  = TagAddrWidth'(sram_tag_bit_addr >>
+  assign sram_tag_word_addr    = TagAddrWidth'(sram_tag_bit_addr >>
                                $clog2(top_pkg::AxiDataWidth));
-  assign sram_tag_bit_select = sram_tag_bit_addr[$clog2(top_pkg::AxiDataWidth)-1:0];
+  assign sram_tag_bit_select_d = sram_tag_bit_addr[$clog2(top_pkg::AxiDataWidth)-1:0];
```
raylau1, line 91: Formatting nit: Align `$clog2` with `TagBitAddrWidth'` and `TagAddrWidth'`

raylau1 (commented):
One formatting nit from my side. Also for me the `uart_smoke_cheri` DV test is timing out with the error:
```
UVM_FATAL @         200000000000: (uvm_phase.svh:1521) [PH_TIMEOUT] Explicit timeout of 200000000000 hit, indicating a probable testbench issue
```

hw/top_chip/rtl/axi_sram.sv
marnovandermaas: Of course, good catch!

marnovandermaas:
Marking this as draft because uart_smoke_cheri is failing.

marnovandermaas:
`uart_smoke_cheri` is now working in DV as well. I came across a bug in the frontend fixed here: https://github.com/lowRISC/cva6-cheri/pull/5

raylau1 (approved):
LGTM!

marnovandermaas:
I also checked that OpenSBI test passes:
```
Boot ROM: v00.01.00
No valid slot found, default to DRAM

Jumping to: 0x80000000

OpenSBI v1.5
   ____                    _____ ____ _____
  / __ \                  / ____|  _ \_   _|
 | |  | |_ __   ___ _ __ | (___ | |_) || |
 | |  | | '_ \ / _ \ '_ \ \___ \|  _ < | |
 | |__| | |_) |  __/ | | |____) | |_) || |_
  \____/| .__/ \___|_| |_|_____/|____/_____|
        | |
        |_|

Platform Name             : lowrisc,mocha
Platform Features         : medeleg
Platform HART Count       : 1
Platform IPI Device       : lowrisc_plicsw
Platform Timer Device     : lowrisc_rvtimer @ 50000000Hz
Platform Console Device   : lowrisc_opentitan_uart
Platform HSM Device       : ---
Platform PMU Device       : ---
Platform Reboot Device    : ---
Platform Shutdown Device  : ---
Platform Suspend Device   : ---
Platform CPPC Device      : ---
Firmware Base             : 0x80000000
Firmware Size             : 323 KB
Firmware RW Offset        : 0x40000
Firmware RW Size          : 67 KB
Firmware Heap Offset      : 0x48000
Firmware Heap Size        : 35 KB (total), 2 KB (reserved), 9 KB (used), 23 KB (free)
Firmware Scratch Size     : 4096 B (total), 624 B (used), 3472 B (free)
Runtime SBI Version       : 2.0

Domain0 Name              : root
Domain0 Boot HART         : 0
Domain0 HARTs             : 0*
Domain0 Region00          : 0x0000000044000000-0x00000000440001ff M: (I,R,W) S/U: ()
Domain0 Region01          : 0x000000004c000000-0x000000004c003fff M: (I,R,W) S/U: ()
Domain0 Region02          : 0x0000000080040000-0x000000008005ffff M: (R,W) S/U: ()
Domain0 Region03          : 0x0000000080000000-0x000000008003ffff M: (R,X) S/U: ()
Domain0 Region04          : 0x0000000048000000-0x000000004bffffff M: (I,R,W) S/U: (R,W)
Domain0 Region05          : 0x0000000000000000-0x0ffffffffffffffff M: () S/U: (R,W,X)
Domain0 Next Address      : 0x0000000080200000
Domain0 Next Arg1         : 0x0000000082200000
Domain0 Next Mode         : S-mode
Domain0 SysReset          : yes
Domain0 SysSuspend        : yes

Boot HART ID              : 0
Boot HART Domain          : root
Boot HART Priv Version    : v1.12
Boot HART Base ISA        : rv64imacb
Boot HART ISA Extensions  : zihpm,zcheripurecap,zcherihybrid
Boot HART PMP Count       : 8
Boot HART PMP Granularity : 3 bits
Boot HART PMP Address Bits: 54
Boot HART MHPM Info       : 6 (0x000001f8)
Boot HART Debug Triggers  : 0 triggers
Boot HART MIDELEG         : 0x0000000000000222
Boot HART MEDELEG         : 0x000000001000b109
TEST RESULT: PASSED
```

PR #521, opened 2026-05-05
[I2C, sw] Move the write to CTRL.ENABLEHOST out of i2c_init

KinzaQamar:
We would use the same function for a device test as well.


elliotb-lowrisc (approved):

martin-velay (approved):
This is required for further dev and doesn't seem to break anything. Thanks Kinza

PR #535, opened 2026-05-07
[I2C, hal, SW] Restructured and refactored the functions

### LINKED ISSUE #559: I2C SW modifications
All the refactoring / restructuring required to extend the existent functionality. For example, 
* Currently SW supports a single byte reads and writes. Extent it to do multiple bytes reads and write
* I2C timing for SCL high cyles needs to satisfy some constrains

KinzaQamar:
Taking the transfer checking part out from the i2c_read_byte() and i2c_write_byte() functions. It is needed when you want to controller wants to read / write multiple bytes.

Merge #545 first as a dependency on this PR

### DIFF
--- a/sw/device/examples/i2c.c
+++ b/sw/device/examples/i2c.c
@@ -29,9 +29,12 @@ int main(void)
         timer_busy_sleep_us(timer, 1000u);
 
         // Read current temperature from an AS6212 I^2C-bus sensor and print the value
-        if (i2c_write_byte(i2c, 0x48u, 0u)) { // select TVAL reg; also a presence check
-            uint16_t sensor_reading = i2c_read_byte(i2c, 0x48u); // read TVAl reg
-            if (sensor_reading != 0xFF) { // only print if we get a non-error value
+        i2c_write_byte(i2c, 0x48u, 0u);
+        if (i2c_wait_write_finish(i2c)) { // select TVAL reg; also a presence check
+            i2c_read_byte(i2c, 0x48u);
+            if (i2c_wait_read_finish(i2c)) {
+                i2c_rdata rdata_reg = VOLATILE_READ(i2c->rdata);
+                uint16_t sensor_reading = rdata_reg.rdata; // read TVAl reg
                 uprintf(uart, "Temperature: 0x%x degC\n",
                         (sensor_reading << 1)); // no decimal printf
             }
--- a/sw/device/examples/i2c.c
+++ b/sw/device/examples/i2c.c
@@ -28,10 +28,11 @@ int main(void)
     while (true) {
         timer_busy_sleep_us(timer, 1000u);
 
+        uint8_t w_data = 0;
         // Read current temperature from an AS6212 I^2C-bus sensor and print the value
-        i2c_write_byte(i2c, 0x48u, 0u);
+        i2c_write_bytes(i2c, 0x48u, &w_data, 1);
         if (i2c_wait_write_finish(i2c)) { // select TVAL reg; also a presence check
-            i2c_read_byte(i2c, 0x48u);
+            i2c_read_bytes(i2c, 0x48u, 1);
             if (i2c_wait_read_finish(i2c)) {
                 uint16_t sensor_reading = i2c_rdata_byte(i2c); // read TVAl reg
                 uprintf(uart, "Temperature: 0x%x degC\n",
--- a/sw/device/examples/i2c.c
+++ b/sw/device/examples/i2c.c
@@ -21,7 +21,7 @@ int main(void)
     timer_enable_write(timer, true);
 
     // Enable I2C in controller mode
-    enable_controller_mode(i2c);
+    i2c_enable_controller_mode(i2c);
 
     uprintf(uart, "Hello i2c Mocha!\n");
 
--- a/sw/device/examples/i2c.c
+++ b/sw/device/examples/i2c.c
@@ -33,8 +33,7 @@ int main(void)
         if (i2c_wait_write_finish(i2c)) { // select TVAL reg; also a presence check
             i2c_read_byte(i2c, 0x48u);
             if (i2c_wait_read_finish(i2c)) {
-                i2c_rdata rdata_reg = VOLATILE_READ(i2c->rdata);
-                uint16_t sensor_reading = rdata_reg.rdata; // read TVAl reg
+                uint16_t sensor_reading = i2c_rdata_byte(i2c); // read TVAl reg
                 uprintf(uart, "Temperature: 0x%x degC\n",
                         (sensor_reading << 1)); // no decimal printf
             }
--- a/sw/device/lib/hal/i2c.c
+++ b/sw/device/lib/hal/i2c.c
@@ -60,7 +60,7 @@ void i2c_init(i2c_t i2c)
     VOLATILE_WRITE(i2c->timing4, t4_reg);
 }
 
-bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
+void i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
 {
     // Reset FMT FIFO (because we currently don't clean-up after errors)
     i2c_fifo_ctrl fifo_ctrl_reg = { .fmtrst = 1u };
@@ -79,24 +79,9 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
     fdata_reg.start = 0;
     fdata_reg.stop = 1u;
     VOLATILE_WRITE(i2c->fdata, fdata_reg);
-
-    // Wait for transaction to complete and report simple succeed/fail
-    for (uint32_t ii = 0; ii < 10000000ul /*arbitrary number*/; ii++) {
-        i2c_intr i2c_intr_state_reg = VOLATILE_READ(i2c->intr_state);
-        if (i2c_intr_state_reg & i2c_intr_controller_halt) {
-            return false; // transaction failed
-        }
-        if (i2c_intr_state_reg & i2c_intr_cmd_complete) {
-            i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
-            if (i2c_status_reg & i2c_status_fmtempty) {
-                return true; // transaction succeeded
-            }
-        }
-    }
-    return false; // timeout
 }
 
-uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
+void i2c_read_byte(i2c_t i2c, uint8_t addr)
 {
     // Reset FMT FIFO (because we currently don't clean-up after errors)
     i2c_fifo_ctrl fifo_ctrl_reg = { .fmtrst = 1u };
@@ -116,21 +101,38 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
     fdata_reg.start = 0;
     fdata_reg.stop = 1u;
     VOLATILE_WRITE(i2c->fdata, fdata_reg);
+}
 
-    // Wait for transaction to complete and return either read data or 0xFF
-    for (uint32_t ii = 0; ii < 10000000ul /*arbitrary number*/; ii++) {
+bool i2c_wait_write_finish(i2c_t i2c)
+{
+    // Wait for transaction to complete and report simple succeed / fail
+    while (true) {
         i2c_intr i2c_intr_state_reg = VOLATILE_READ(i2c->intr_state);
         if (i2c_intr_state_reg & i2c_intr_controller_halt) {
-            return 0xFF; // transaction failed
+            return false; // Transaction failed
+        }
+        if (i2c_intr_state_reg & i2c_intr_cmd_complete) {
+            i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
+            if (i2c_status_reg & i2c_status_fmtempty) {
+                return true; // Transaction succeeded
+            }
+        }
+    }
+}
+
+bool i2c_wait_read_finish(i2c_t i2c)
+{
+    // Wait for transaction to complete and report simple succeed / fail
+    while (true) {
+        i2c_intr i2c_intr_state_reg = VOLATILE_READ(i2c->intr_state);
+        if (i2c_intr_state_reg & i2c_intr_controller_halt) {
+            return false; // Transaction failed
         }
         i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
         if (i2c_status_reg & i2c_status_fmtempty) {
-            // transaction succeeded, return read data
-            i2c_rdata rdata_reg = VOLATILE_READ(i2c->rdata);
-            return rdata_reg.rdata;
+            return true;
         }
     }
-    return 0xFF; // timeout
 }
 
 void enable_controller_mode(i2c_t i2c)
--- a/sw/device/lib/hal/i2c.c
+++ b/sw/device/lib/hal/i2c.c
@@ -77,8 +77,13 @@ void i2c_write_bytes(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_b
 
     fdata_reg.start = 0;
 
-    // Send all data bytes; assert STOP only on the last byte
     for (uint8_t i = 0; i < num_bytes; i++) {
+        // Check the overflow condition first before writing to the FMT FIFO by waiting until FMT
+        // FIFO has some space
+        while (VOLATILE_READ(i2c->status) & i2c_status_fmtfull) {
+        }
+
+        // Send all data bytes; assert STOP only on the last byte
         fdata_reg.fbyte = data[i];
         if (i == (num_bytes - 1u)) {
             fdata_reg.stop = 1u;
--- a/sw/device/lib/hal/i2c.c
+++ b/sw/device/lib/hal/i2c.c
@@ -60,7 +60,7 @@ void i2c_init(i2c_t i2c)
     VOLATILE_WRITE(i2c->timing4, t4_reg);
 }
 
-void i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
+void i2c_write_bytes(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_bytes)
 {
     // Reset the FMT FIFO as a precautionary step in case something goes wrong when controller's FSM
     // is halted and the SW didn't manage to clear the FIFO during that scenario.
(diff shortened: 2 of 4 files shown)


sw/device/examples/i2c.c
```diff
@@ -26,9 +26,11 @@ int main(void)
...
 
         // Read current temperature from an AS6212 I^2C-bus sensor and print the value
-        if (i2c_write_byte(i2c, 0x48u, 0u)) { // select TVAL reg; also a presence check
-            uint16_t sensor_reading = i2c_read_byte(i2c, 0x48u); // read TVAl reg
-            if (sensor_reading != 0xFF) { // only print if we get a non-error value
+        i2c_write_byte(i2c, 0x48u, 0u);
+        if (check_wr_xfer_status(i2c)) { // select TVAL reg; also a presence check
+            i2c_read_byte(i2c, 0x48u);
+            if (check_rd_xfer_status(i2c)) {
+                uint16_t sensor_reading = DEV_READ(i2c + I2C_RDATA_REG); // read TVAl reg
```
engdoreis: Please keep mmio access inside the hal

sw/device/lib/hal/i2c.c
```diff
@@ -49,51 +49,65 @@ void i2c_init(i2c_t i2c)
                                          (rnd_up_div(4000, SYSCLK_NS) << I2C_TIMING4_TSU_STO));
 }
 
-bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
+void i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
 {
+    uint32_t wr_xfer; // Stores write transfer information
```
engdoreis: since you are doing a lot of re-write, to avoid having to review this again.  You should  use the autogenerated rdl headers in https://github.com/lowRISC/mocha/blob/main/sw/device/lib/hal/autogen/i2c.h

Check the hal/uart for a reference.

sw/device/examples/i2c.c
KinzaQamar: I don't understand what exactly is wrong on L34?

sw/device/lib/hal/i2c.c
KinzaQamar: Yeah, we discuss this in person and the plan is to do a follow-up PR

sw/device/lib/hal/i2c.c
```diff
@@ -60,9 +60,10 @@ void i2c_init(i2c_t i2c)
     VOLATILE_WRITE(i2c->timing4, t4_reg);
 }
 
-bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
+void i2c_write_n_bytes(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_wr_bytes)
 {
-    // Reset FMT FIFO (because we currently don't clean-up after errors)
+    // Reset the FMT fifo as a precautionary step in case something goes wrong when controller's FSM
+    // was halted and the SW didn't manage to clear the fifo during that scenario.
```
elliotb-lowrisc: Nit: mixture of past and future tense ("goes wrong", "was halted")

(Also, I'd use "FIFO" rather than "fifo" in a sentence as it is an acronym

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,22 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
     }
-    return false; // timeout
 }
 
-uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
+void i2c_read_n_bytes(i2c_t i2c, uint8_t addr, uint8_t num_rd_bytes)
 {
-    // Reset FMT FIFO (because we currently don't clean-up after errors)
+    // Reset the FMT fifo as a precautionary step in case something goes wrong when controller's FSM
+    // was halted and the SW didn't manage to clear the fifo during that scenario.
```
elliotb-lowrisc: Nit: mix of tenses again

sw/device/lib/hal/i2c.c
```diff
@@ -112,28 +104,67 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
...
+                return true; // transaction succeeded
+            }
+        }
+    }
+    return false; // Timeout
+}
+
+bool check_rd_xfer_status(i2c_t i2c)
+{
     // Wait for transaction to complete and return either read data or 0xFF
```
elliotb-lowrisc: Code comment needs updating

KinzaQamar: Good point. 

Done now, thanks!

KinzaQamar:
Tested `examples/i2c.c` on an FPGA server:

image

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,22 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
-            if (i2c_status_reg & i2c_status_fmtempty) {
-                return true; // transaction succeeded
-            }
+    // Send all data bytes; assert STOP only on the last byte
+    for (uint8_t i = 0; i < num_wr_bytes; i++) {
+        fdata_reg.fbyte = data[i];
+        if (i == (num_wr_bytes - 1u)) {
+            fdata_reg.stop = 1u;
         }
+        VOLATILE_WRITE(i2c->fdata, fdata_reg);
```
engdoreis, line 91: you should check that the fifo is not full on every iteration, otherwise it would overflow when num_wr_bytes is large.

sw/device/lib/hal/i2c.c
```diff
@@ -112,28 +104,67 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
...
+    fdata_reg.fbyte = num_rd_bytes; // If readb = 1 then fbyte contains the number of bytes to read
     fdata_reg.start = 0;
     fdata_reg.stop = 1u;
     VOLATILE_WRITE(i2c->fdata, fdata_reg);
+}
+
+bool check_wr_xfer_status(i2c_t i2c)
+{
+    // Wait for transaction to complete and report simple succeed / fail
+    for (uint32_t ii = 0; ii < 10000000ul /*arbitrary number*/; ii++) {
```
engdoreis: why we need to check the timeout? The arbitrary number of iterations is brittle. Since this code is only used for testing, and tests already have timeout mechanims. I would replace the `for` loop for a `while (true)`

Same for the function below.

sw/device/lib/hal/i2c.c
```diff
@@ -112,28 +104,67 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
...
     // Send stop bit, read bit and number of bytes to read
     fdata_reg.readb = 1u;
-    fdata_reg.fbyte = 1u; // If readb = 1 then fbyte contains the number of bytes to read
+    fdata_reg.fbyte = num_rd_bytes; // If readb = 1 then fbyte contains the number of bytes to read
     fdata_reg.start = 0;
     fdata_reg.stop = 1u;
     VOLATILE_WRITE(i2c->fdata, fdata_reg);
+}
+
+bool check_wr_xfer_status(i2c_t i2c)
```
engdoreis: The check name sugests that there's no loop.
```suggestion
bool wait_for_completion(i2c_t i2c)
```
Same for the function below.

sw/device/lib/hal/i2c.c
KinzaQamar: This is going to be `UVM agent` receiving these bytes. It is capable of receiving any number of bytes. 

I was thinking about checking `FMTFULL` before sending a transfer but since we always write `1` to `fmtrst` (which is going to reset the FMT FIFO), this is not needed. About checking `RXFULL`, this should be a test specific thing.

elliotb-lowrisc: The UVM agent may be fast, but the `i2c` block itself might not be. If the processor writes data into the FMT FIFO faster than the I^2C hardware can transmit it, then all it takes is a sufficiently large data transfer for data to be lost when the processor eventually writes to an already-full FMT FIFO.

KinzaQamar: > you should check that the fifo is not full on every iteration, otherwise it would overflow when num_wr_bytes is large.

Oh, sorry. I misunderstood your comment thinking as if you are talking about `target's RX FIFO` overflow.

Thanks for clarification @elliotb-lowrisc

engdoreis (approved):
LGTM

KinzaQamar:
Hi @engdoreis,

I've addresses your [comment](https://github.com/lowRISC/mocha/pull/535#discussion_r3235468654). Meanwhile, testing the change I found out that a loop back test can't happen when the bytes we want to read are greater than FIFO_DEPTH

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,35 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
-    // Wait for transaction to complete and report simple succeed/fail
-    for (uint32_t ii = 0; ii < 10000000ul /*arbitrary number*/; ii++) {
-        i2c_intr i2c_intr_state_reg = VOLATILE_READ(i2c->intr_state);
-        if (i2c_intr_state_reg & i2c_intr_controller_halt) {
-            return false; // transaction failed
+    // Send all data bytes; assert STOP only on the last byte
+    for (uint8_t i = 0; i < num_wr_bytes; i++) {
+        // The current FIFO depth is 64. So, raise the flag whenever the byte number is a multiple
+        // of 64.
+        bool overflow = ((i % FIFO_DEPTH) == 0);
```
engdoreis: This is clever from the performance point of view, but it's not generic enough, if the FIFO is partially filled before this function is called, then we could have an overflow.
Since performance is not problem here, you should always check the for overflow before any data is put into the FIFO.

KinzaQamar: That's not going to be partially filled since at the start of the function we reset the FMT FIFO [changes#diff-b4dbf3ddde1f437b44c71b9905997276ad2b88c790eb7c8aeef84dd6a3558134L66](https://github.com/lowRISC/mocha/pull/535/changes#diff-b4dbf3ddde1f437b44c71b9905997276ad2b88c790eb7c8aeef84dd6a3558134L66).

According to programmer's guide, it ask to check if the FMT FIFO is empty after every read / write transfer to confirm that the transfer is succesful.

I have an extended question though. Do we really need the precautionary step of clearing FMT FIFO before every read / write now that we are clearing it on errors? 

You comment is valid when someone is not properly following the programmer's guide and we remove the precautionary step at the start of `i2c_write_n_bytes` and `i2c_read_n_bytes`.

sw/device/lib/hal/i2c.h
```diff
@@ -90,17 +90,17 @@
...
 
-// Transmits a single byte to the target
-void i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data);
+// Transmits multiple bytes to the target
+void i2c_write_n_bytes(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num_wr_bytes);
 
-// Receive a single byte from the target
-void i2c_read_byte(i2c_t i2c, uint8_t addr);
+// Receive n bytes from the target
+void i2c_read_n_bytes(i2c_t i2c, uint8_t addr, uint8_t num_rd_bytes);
```
ziuziakowska: Would prefer these to be called `i2c_write_bytes` and `i2c_read_bytes`, and the length parameter should just be called `len`.

sw/device/lib/hal/i2c.h
```diff
@@ -90,17 +90,17 @@
...
 // Check if the write was successful
-bool check_rd_xfer_status(i2c_t i2c);
+bool wait_rd_xfer_status(i2c_t i2c);
 
 // checks if the read was successful
-bool check_wr_xfer_status(i2c_t i2c);
+bool wait_wr_xfer_status(i2c_t i2c);
 
 // Enable I2C in controller mode
 void enable_controller_mode(i2c_t i2c);
```
ziuziakowska: These functions should have the `i2c_` prefix, and the abbreviations (rd, wr, xfer) should be expanded.

sw/device/lib/hal/i2c.c
```diff
@@ -79,10 +79,23 @@ void i2c_write_n_bytes(i2c_t i2c, uint8_t addr, const uint8_t *data, uint8_t num
...
+
+        // Check the overflow condition on every byte which is a multiple of 64.
+        if (overflow) {
+            i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
+
+            // Wait until FMT FIFO has some space
+            while (i2c_status_reg & i2c_status_fmtfull) {
+                i2c_status_reg = VOLATILE_READ(i2c->status);
+            }
+        }
```
ziuziakowska: I think there is a bug here - it could be that after writing 64 bytes, the FIFO isn't full but is also *not empty*, so another 64 bytes would be written before checking the FIFO status again, which would overflow it. It's much more robust and simple to check whether the FIFO is full before writing each byte.

sw/device/lib/hal/i2c.h
KinzaQamar: I wanted an `n` in b/w to indicate that the function can read / write one byte as well.

KinzaQamar: Hmm, I think I'd go with your suggestion. As `n_bytes` and only `bytes` are same meaning

KinzaQamar: > and the length parameter should just be called len

I wanted to call it with a `_bytes` suffix to be explicit that these are the number of bytes to read and write. I can change it to `num_bytes` or `len_bytes`. Does that look sensible to you?

sw/device/lib/hal/i2c.c
KinzaQamar: Hmm, that's interesting. I wanted to only check an overflow when there is a chance of overflow. I2C controller is fast enough to drain the FIFO so we will never get it full. But your comment is right that when the SW is fast and controller is slow, this check will lead to data loss. Thanks @ziuziakowska

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,30 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
 }
 
-uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
+void i2c_read_bytes(i2c_t i2c, uint8_t addr, uint8_t len_bytes)
 {
-    // Reset FMT FIFO (because we currently don't clean-up after errors)
+    // Reset the FMT FIFO as a precautionary step in case something goes wrong when controller's FSM
+    // is halted and the SW didn't manage to clear the FIFO during that scenario.
     i2c_fifo_ctrl fifo_ctrl_reg = { .fmtrst = 1u };
     VOLATILE_WRITE(i2c->fifo_ctrl, fifo_ctrl_reg);
```
KinzaQamar, line 100: @elliotb-lowrisc Do we still need to keep this precautionary step before the start of every read and write no that we are clearing on errors?

sw/device/lib/hal/i2c.c
KinzaQamar: Addressed it as part of [r3280355755](https://github.com/lowRISC/mocha/pull/535#discussion_r3280355755) comment

KinzaQamar:
Comments addressed !

sw/device/lib/hal/i2c.c
elliotb-lowrisc: Sorry I missed this for so long. I don't know, but I would think that it can't hurt to leave it in there (minor performance impact)

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,30 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
-        if (i2c_intr_state_reg & i2c_intr_controller_halt) {
-            return false; // transaction failed
+    for (uint8_t i = 0; i < len_bytes; i++) {
+        // Check the overflow condition before writing to the FMT FIFO.
+        i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
+
+        // Wait until FMT FIFO has some space
+        while (i2c_status_reg & i2c_status_fmtfull) {
+            i2c_status_reg = VOLATILE_READ(i2c->status);
         }
```
engdoreis:
```suggestion
         // Wait until FMT FIFO has some space
        while (VOLATILE_READ(i2c->status) & i2c_status_fmtfull) {}
```

sw/device/lib/hal/i2c.c
```diff
@@ -112,28 +112,67 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
...
+            return false; // Transaction failed
+        }
+        if (i2c_intr_state_reg & i2c_intr_cmd_complete) {
+            i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
+            if (i2c_status_reg & i2c_status_fmtempty) {
+                return true; // Transaction succeeded
+            }
+        }
+    }
+    return false; // Timeout
```
engdoreis: Delete this line, it's unreached.

sw/device/lib/hal/i2c.c
```diff
@@ -112,28 +112,67 @@ uint8_t i2c_read_byte(i2c_t i2c, uint8_t addr)
...
         i2c_status i2c_status_reg = VOLATILE_READ(i2c->status);
         if (i2c_status_reg & i2c_status_fmtempty) {
-            // transaction succeeded, return read data
-            i2c_rdata rdata_reg = VOLATILE_READ(i2c->rdata);
-            return rdata_reg.rdata;
+            return true;
         }
     }
-    return 0xFF; // timeout
+    return false; // Timeout
```
engdoreis: Delete this line, it's unreached.

KinzaQamar: Thanks

sw/device/lib/hal/i2c.c
KinzaQamar: Yes, true. I didn't notice this, Now done, thanks!

sw/device/lib/hal/i2c.c
```diff
@@ -74,31 +75,28 @@ bool i2c_write_byte(i2c_t i2c, uint8_t addr, uint8_t data)
...
-    VOLATILE_WRITE(i2c->fdata, fdata_reg);
 
-    // Wait for transaction to complete and report simple succeed/fail
-    for (uint32_t ii = 0; ii < 10000000ul /*arbitrary number*/; ii++) {
-        i2c_intr i2c_intr_state_reg = VOLATILE_READ(i2c->intr_state);
-        if (i2c_intr_state_reg & i2c_intr_controller_halt) {
-            return false; // transaction failed
+    for (uint8_t i = 0; i < num_bytes; i++) {
+        // Check the overflow condition before writing to the FMT FIFO.
+        //
```
engdoreis: Delete empty comment line?

engdoreis (approved):

sw/device/lib/hal/i2c.c
KinzaQamar: No, that was intentional. I wanted to make it separate comments within the same context

KinzaQamar: I've just push a change which connects both comments into a single comment

engdoreis (approved):

PR #548, opened 2026-05-12
[doc] SRAM specification added

marnovandermaas:
This adds an initial specification of the SRAM block.

### DIFF
--- a/doc/ref/arch.md
+++ b/doc/ref/arch.md
@@ -38,3 +38,39 @@ In terms of output, the top-level will need output signals:
 - Key to provide an AES engine outside of the secure enclave with the memory encryption key.
 - AXI manager port to interact with the rest of the chip.
 
+## SRAM specification
+
+The static random-access memory (SRAM) in CHERI Mocha is mainly used as the stack and heap for the boot firmware that lives in the read-only memory (ROM).
+However, it should also be possible to execute from SRAM.
+Once code starts executing from dynamic random-access memory (DRAM), we don't envision using SRAM anymore.
+
+The SRAM block has four ports:
+- Clock input
+- Reset input
+- AXI4 request input from the main SoC sub-system crossbar
+- AXI4 response output back to the main crossbar
+
+Inside the block it translates the AXI4 requests into an SRAM interface that our primitive RAM wrappers use.
+It needs to support AXI4 protocol including:
+- Bursts, where the last signal must be indicated correctly.
+- Response must have the same AXI4 ID as the request
+- Atomic support is *excluded*.
+- The data width is 64 bits.
+- The address range and size of the SRAM are defined in the [memory map](#memory-map). Accesses outside this range must return an error, including if only part of the burst is outside the memory range.
+- Responses must return within a bounded amount of time that may be proportional to the length of the burst.
+- Only aligned 64-bit accesses are allowed.
+
+There needs to be 1 CHERI tag bit per 128-bit aligned region.
+A tag should only be set to 1 by writing a full 128-bit aligned region.
+This 128-bit aligned transaction must be part of a single burst.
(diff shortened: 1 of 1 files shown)


doc/ref/arch.md
```diff
@@ -38,3 +38,33 @@ In terms of output, the top-level will need output signals:
...
+- Response must have the same AXI4 ID as the request
+- The last signal must be indicated correctly.
+- Atomic support is *excluded*.
+- The data width is 64 bits.
+- The address range and size of the SRAM are defined in the [memory map](#memory-map). Accesses outside this range must return an error, including if only part of the burst is outside the memory range.
+- Responses must follow the request with a single cycle of latency.
+
+There needs to be 1 CHERI tag bit per 128-bit aligned region.
+A tag can only be set to 1 by writing a full 128-bit aligned region.
+If a portion of the 128-bit aligned region is written it must clear the tag for the whole region, this includes when a partial write strobe is used.
```
tchilikov-semify: Regarding these two points:

- A tag can only be set to 1 by writing a full 128-bit aligned region.
- If a portion of the 128-bit aligned region is written it must clear the tag for the whole region, this includes when a partial write strobe is used.

Is the core responsible for enforcing these rules, or the SRAM? I.e. If a 128-bit region is already set with a tag, then a partial write comes in, the tag should be cleared. Does the clearing happen by the SRAM, or does the SRAM always blindly accept the tag value passed in by wuser?

doc/ref/arch.md
```diff
@@ -38,3 +38,33 @@ In terms of output, the top-level will need output signals:
...
+The CHERI tag bits are communicated through a single user bit on the AXI4 bus (`wuser` and `ruser` for writes and reads respectively).
+Burst reads from the SRAM must have the appropriate CHERI tags set for each address, so a valid capability must have the user bits set for both of the 64-bit flits it is being sent back, and a mixture of capability and non-capability data is allowed in a burst.
```
tchilikov-semify, line 72: Just for my understanding, regarding point:

- Burst reads from the SRAM must have the appropriate CHERI tags set for each address, so a valid capability must have the user bits set for both of the 64-bit flits it is being sent back, and a mixture of capability and non-capability data is allowed in a burst.

If a 128-bit tag valid field is partially written to, the tag shall be cleared. But if the same 128-bit field is partially read, the partial result shall return the valid flag? Two partial reads of the same 128-bit field will return the same valid flag.

Is this reasoning correct?

doc/ref/arch.md
marnovandermaas: I think you're right, that some of this is the responsibility of the core. The SRAM is mainly responsible for the things that happen sub-64-bit. So if a 32-bit write occurs this must set the tag to zero. We should probably also add a simulation assertion that if it is not a full 64-bit write that wuser cannot be high. Let me try clearing this up in the documentation.

marnovandermaas: On second thought, I think it is reasonable to require that a valid capability is only valid if it is written as part of the same burst. This allows us to only store one bit per 128-bit region instead of having one per 64-bit region.

doc/ref/arch.md
marnovandermaas: Yes, I think 64-bit partial reads from a capability should be marked as valid. But sub-64-bit reads must always have ruser set to zero. I'll try to update the documentation accordingly.

marnovandermaas: Let me know if the new documentation makes this more clear.

doc/ref/arch.md
marnovandermaas: Let me know if the new documentation makes this more clear.

tchilikov-semify:
The updated spec is clear and looks good, approved from my side

martin-velay (approved):
LGTM, thanks for this spec Marno this will help to make proper DV 👍

tchilikov-semify (approved):

PR #565, opened 2026-05-18
[i2c, sw, hal] Added functions to calculate timing params for all supported speeds and correct SCL high time

### LINKED ISSUE #559: I2C SW modifications

KinzaQamar:
First commit takes care that the resultant SCL high cycles must satisfy below equations:
** scl_high_cycles >= 4 (to support correct function in clock streching)
** scl_high_cycles = scl_period_cycles - (rise_cycles + fall_cycles
                     + scl_low_cycles)
                     
Second commit added calculations for each supported timing parameters

### DIFF
--- a/sw/device/lib/hal/i2c.c
+++ b/sw/device/lib/hal/i2c.c
@@ -23,11 +23,9 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
 }
 
 uint16_t i2c_calc_scl_high_cycles(uint16_t rise_cycles, uint16_t fall_cycles,
-                                  uint16_t scl_period_cycles, uint16_t scl_low_cycles)
+                                  uint16_t scl_period_cycles, uint16_t scl_low_cycles,
+                                  uint16_t scl_high_cycles_min)
 {
-    // Calculate the minimum allowable value for SCL high time
-    uint16_t scl_high_cycles_min = rnd_up_div(4000, SYSCLK_NS);
-
     // scl_high_time should be atleast 4 cycles to aid correct clock streching
     scl_high_cycles_min = (scl_high_cycles_min < 4u) ? 4u : scl_high_cycles_min;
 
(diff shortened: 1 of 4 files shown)


sw/device/lib/hal/i2c.c
```diff
@@ -22,6 +22,33 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
     return (uint16_t)result;
 }
 
+static uint16_t calc_scl_high_cycles(uint32_t rise_cycles, uint32_t fall_cycles,
+                                     uint32_t scl_period_cycles, uint32_t scl_low_cycles)
+{
+    // Calculate the minimum allowable value for SCL high time
+    uint16_t scl_high_cycles_min = rnd_up_div(4000, SYSCLK_NS);
```
engdoreis: Where the 4000 comes from? Can you make it a constant variable?

engdoreis (commented):
I think that we should bite the bullet and port the Opentitan dif_i2c to mocha. Starting by the [timing functions](https://github.com/lowRISC/opentitan/blob/7791207ad2a60ca8349fa1a801f3a5c639be21db/sw/device/lib/dif/dif_i2c.c#L139).

KinzaQamar:
> I think that we should bite the bullet and port the Opentitan dif_i2c to mocha. Starting by the [timing functions](https://github.com/lowRISC/opentitan/blob/7791207ad2a60ca8349fa1a801f3a5c639be21db/sw/device/lib/dif/dif_i2c.c#L139).

You're right. I'll do this in a separate PR

KinzaQamar:
> > I think that we should bite the bullet and port the Opentitan dif_i2c to mocha. Starting by the [timing functions](https://github.com/lowRISC/opentitan/blob/7791207ad2a60ca8349fa1a801f3a5c639be21db/sw/device/lib/dif/dif_i2c.c#L139).
> 
> You're right. I'll do this in a separate PR

I've done it as part of this PR

sw/device/lib/hal/i2c.h
```diff
@@ -88,7 +88,23 @@
...
+    uint32_t scl_low_cycles;
+    uint32_t scl_high_cycles;
+    uint32_t scl_period_cycles;
+    uint32_t setup_start_cycles;
+    uint32_t hold_start_cycles;
+    uint32_t setup_data_cycles;
+    uint32_t hold_data_cycles;
+    uint32_t setup_stop_cycles;
+    uint32_t bus_free_time_cycles;
+} i2c_timing_params_cycles_t;
```
ziuziakowska: I would prefer that `enums` and `structs` are not typedef-ed out (with some exceptions, e.g volatile-qualified pointers for devices like `i2c_t`)

sw/device/lib/hal/i2c.h
```diff
@@ -88,7 +88,23 @@
 #define I2C_RISE_NS (450)
 #define I2C_FALL_NS (120)
 
-void i2c_init(i2c_t i2c);
+typedef enum { standard_mode, fast_mode, fast_plus_mode } i2c_speed_t;
```
ziuziakowska: Convention in the auto-generated headers is to have the enum variants be prefixed with the name of the enum they are part of, so this type should be `i2c_speed_mode` and the variants should be `i2c_speed_mode_standard`, `i2c_speed_mode_fast`, and `i2c_speed_mode_fast_plus`. These should also be one variant per-line.

sw/device/lib/hal/i2c.h
ziuziakowska: I think `cycles` in the type name is redundant as the units are in the struct member names already. Perhaps this should be `i2c_timing_parameters`?

sw/device/lib/hal/i2c.c
```diff
@@ -49,37 +47,82 @@ uint16_t calc_scl_high_cycles(uint32_t rise_cycles, uint32_t fall_cycles,
     return (uint16_t)(scl_high_cycles);
 }
 
-void i2c_init(i2c_t i2c)
+// Calculate the minimum allowable value for each timing parameter taken from the NXP I^2C
+// specification "UM10204" Table 10 (rev. 6) / Table 11 (rev. 7).
+//
+// The values for Rise and Fall times for Fast mode are taken as spec minimum. For Fast plus mode,
+// the values are taken from OT's i2c_host_tx_rx_test.c test.
+i2c_timing_params_cycles_t compute_min_timing_params(i2c_speed_t speed)
```
ziuziakowska: This should be called `minimum_timing_parameters` I think

sw/device/lib/hal/i2c.c
```diff
@@ -22,6 +22,33 @@ static uint16_t rnd_up_div(uint32_t a, uint32_t b)
     return (uint16_t)result;
 }
 
+uint16_t calc_scl_high_cycles(uint32_t rise_cycles, uint32_t fall_cycles,
+                              uint32_t scl_period_cycles, uint32_t scl_low_cycles)
```
ziuziakowska: Should these not all be `uint16_t`?

ziuziakowska (commented):
Just some style nits.

sw/device/lib/hal/i2c.h
KinzaQamar: Just naming it `i2c_timing_paramaters` will get misunderstood as timing parameters in `ns`

KinzaQamar: These typedefs are going to be used as function arguments and  return types. Isn't it nice to make them a type and use them wherever required?

sw/device/lib/hal/i2c.c
KinzaQamar: Yes, thanks!

KinzaQamar:
Hi @ziuziakowska 

Thanks for your review. I've addressed your comments and left some questions.

sw/device/lib/hal/i2c.h
engdoreis: I ok in keeping the cycles in the struct name, but cycles in the fields name is redundat.
Looking above when used, cycles appears twice, which just polutes.
```sh
 .tsu_sto = timing_params_cycles.setup_stop_cycles
```

KinzaQamar: I've took the `_cycles` suffix out from the struct name

engdoreis (approved):

sw/device/lib/hal/i2c.h
ziuziakowska: The types would still be available when the header is included, you would just need to prefix them with `enum` or `struct`. I generally prefer not hiding the specifics of a type behind `typedef`s, e.g `boot/*.h` and `hal/mocha_irq.h`

martin-velay:
Since this is a SW related PR and got approved from Douglas, I guess it's fine to merge it

sw/device/lib/hal/i2c.h
martin-velay: Oups, sorry I just noticed your comment @ziuziakowska and I've already pressed the "merge" button!
Shall I create a new PR to implement your requested changes?

PR #567, opened 2026-05-19
Ethernet patches

marnovandermaas:
These patches have been moved to the upstream repository so we can remove these from the repository now.


raylau1 (approved):
LGTM.

PR #387, opened 2026-03-26
Update SPI Device HAL to use new autogen interface

ziuziakowska:
Split off from https://github.com/lowRISC/mocha/pull/299.

~~Depends on https://github.com/lowRISC/mocha/pull/579.~~ Merged.

PR #579, opened 2026-05-21
Various RDL Autogen fixes

### LINKED ISSUE #522: RDL generated header uses 32-bit array type when memwidth is 64
The auto generated RDL header uses uint32_t for memories with memwidth of 64. This causes assertion errors when running clang_tidy.

An example RDL and header with this issue can be found on branch [ethernet_rdl](https://github.com/lowRISC/mocha/tree/ethernet_rdl).

ziuziakowska:
- Add a "none" variant to "flag enum" generated registers. This is useful for representing a "no interrupts" or "no bits" value to pass to `enable_write` or `status_write` to clear the register, needed by https://github.com/lowRISC/mocha/pull/387.
- Support generating registers interspersed with register windows - it was assumed before that all devices have their register windows after their register fields, but this is not true of SPI Host.
- Support generating register windows with 64-bit entries, as needed for Ethernet.

This also adds the autogenerated SPI Host and Ethernet headers.

Closes https://github.com/lowRISC/mocha/issues/522.


util/rdlgenerator.py
```diff
@@ -425,16 +425,31 @@ def emit_device_register_field(device_name: str, reg: dict, type_name: str) -> s
 
 def emit_device_window_field(device_name: str, window: dict) -> str:
     window_name = window["name"].lower()
-    num_u32s = window["entries"]
+    num_entries = window["entries"]
+    mem_width = window["width"]
+    mem_width_bytes = 0
+    type_name = None
+    if mem_width == 32:
```
engdoreis: use [match statement](https://docs.python.org/3/tutorial/controlflow.html#tut-match) or dict mapping?

util/rdlgenerator.py
```diff
@@ -425,16 +425,31 @@ def emit_device_register_field(device_name: str, reg: dict, type_name: str) -> s
...
+        raise ValueError(
+            f"Register window '{window_name}' in device '{device_name}' "
+            "has entries that are not 32-bits or 64-bits wide"
+        )
+
     offset = window["offset"]
-    end = offset + window["size"] - 4
-    offset_string = f"{hex(offset)}-{hex(end)}"
+    end = offset + window["size"] - mem_width_bytes
+    offset_string = hex(offset) if num_entries == 1 else f"{hex(offset)}-{hex(end)}"
```
engdoreis: Nit:
```suggestion
    offset_string = hex(offset) if num_entries == 1 else f"{offset:x}-{end:x}"
```

util/rdlgenerator.py
```diff
@@ -451,43 +466,60 @@ def emit_device_struct_declaration(
...
     """
     Emit the declaration of the device's memory layout. The fields of this structure are the
-    device's registers, with necessary padding fields around them.
+    device's registers and windows, with necessary padding fields around them.
     """
-    typed_registers_ascending_by_offset = sorted(
-        typed_registers, key=lambda reg: min(reg[0]["offsets"])
+    # all the struct declaration fields. Tag each with a boolean value representing whether
+    # it is a register or not (a window).
+    declaration_fields = [(True, register) for register in typed_registers] + [
```
engdoreis: The tuple isn't great for readability. The json file has a field "type" can be used instead of the tuple. Or you could simply check if the `offset` field  is an array

engdoreis (commented):
Thanks for this important fix, just three comments to be addressed.

util/rdlgenerator.py
```diff
@@ -266,10 +266,12 @@ def register_is_flag_enum(reg: dict) -> bool:
 def emit_register_flag_enum(device_name: str, reg: dict) -> str:
     """
     Emit register as a flag enum, see 'register_is_flag_enum' predicate.
+    We additionally emit a 'none' variant with a value of 0.
     """
     reg_name = reg["name"].lower()
     fully_qualified_type_name = "_".join([device_name, reg_name])
     enum_variants = []
+    enum_variants.append(f"{'_'.join([device_name, reg_name, 'none'])} = (0),")
```
AlexJones0: Is there a reason we emit `(0)`? This is completely subjective, but I'd much prefer just `0`.

util/rdlgenerator.py
AlexJones0: Alternatively, in this case I'd actually prefer:

```py
    if mem_width not in (32, 64):
        raise ValueError(
            f"Register window '{window_name}' in device '{device_name}' "
            "has entries that are not 32-bits or 64-bits wide"
        )

    mem_width_bytes = mem_width // 8
    type_name = f"uint{mem_width}_t"
```

util/rdlgenerator.py
```diff
@@ -448,43 +466,60 @@ def emit_device_struct_declaration(
...
-        typed_registers, key=lambda reg: min(reg[0]["offsets"])
+    # all the struct declaration fields. Tag each with a boolean value representing whether
+    # it is a register or not (a window).
+    declaration_fields = [(True, register) for register in typed_registers] + [
+        (False, window) for window in windows
+    ]
+
+    # sort the struct fields in ascending order by offset. the specific keys differ based
+    # on whether it is a register or a window.
+    fields_asceding_by_offset = sorted(
```
AlexJones0: Typo:
```suggestion
    fields_ascending_by_offset = sorted(
```

util/rdlgenerator.py
AlexJones0: I mentioned this on previous PR(s) but these would really benefit from being a bit more object-oriented with either dataclasses or Pydantic models, it would help to solve this readability a bit. Albeit that's obviously a larger refactoring task.

For now, if nothing else is workable, this might be more readable as its own dict e.g.: `{"is_window": False, "field": register}`, even if a bit less efficient.

AlexJones0 (commented):
I've not checked the RDL, only the SW changes.

util/rdlgenerator.py
ziuziakowska: That's a lot nicer, thanks!

util/rdlgenerator.py
ziuziakowska: Didn't know about this feature, looks much less noisy

engdoreis: Very useful format specifiers, it's my hex and bin calculator :smile: 
```sh
 python -c "print(f'{0x20+0b11:x}')"
 python -c "print(f'{0x20+0b11:b}')"
```

engdoreis (approved):
LGTM. Thanks for the changes

raylau1 (approved):
Works on my side.

PR #582, opened 2026-05-26
[rtl,sw] Fix tag cache eviction write logic

### LINKED ISSUE #528: OpenSBI SBI call causing infinite loop
See: https://github.com/ziuziakowska/u-boot/commit/e2740585d8c324c2ae5f54f0af0380c91e457fb1

OpenSBI enters an infinite loop when the trap handler uses an invalid capability early in the handler code. I haven't yet ruled out whether this is a toolchain bug or a hardware issue.

This is quite critical to Operating System support, as SBI calls are used by Linux and BSD to schedule timer interrupts and early debugging. This may also be responsible for Integer mode Linux freezes early in the boot process.

raylau1:
This PR moves the overflow flag logic added in the AXI LLC to the common_cells counter. This fixes the tag cache eviction write bug which caused the OpenSBI tag fault.

The tag controller tag test is also updated to check evicted tags.

Closes #528.

### DIFF
--- a/hw/ip/common_cells/rtl/counter.sv
+++ b/hw/ip/common_cells/rtl/counter.sv
@@ -33,6 +33,13 @@ module counter #(
     .commit_i           (1'b1),
     .cnt_o              (q_o),
     .cnt_after_commit_o ( ),
-    .err_o              (overflow_o)
+    .err_o              ( )
   );
+
+  // Generate sticky overflow flag as prim_count is a saturating counter
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if      (!rst_ni)             overflow_o <= 1'b0;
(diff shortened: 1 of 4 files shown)


marnovandermaas (commented):
I assume this was an issue because of the use in this file: `src/eviction_refill/axi_llc_w_master.sv` ?
Maybe good to mention that in the commit message. I'll wait on approving this until @ziuziakowska confirms that the issue is resolved.

ziuziakowska:
I tried running OpenSBI + U-boot both in CHERI Mode with this, and they both worked perfectly!

marnovandermaas (approved):
Hardware changes look good and also after confirming the software issue is resolved.

PR #586, opened 2026-05-27
[doc] Move badge domain from dashboards to dashboard.reports

tye-b:
Soon badges will be migrating from the `dashboards.lowrisc.org` domain to the `dashboard.reports.lowrisc.org`. For now they exist in both but to get ahead of the migration it makes sense to change them now.


marnovandermaas (commented):
You'll need to patch this in hw/vendor/patches/lowrisc_ip/prim

marnovandermaas (approved):
Thanks for the fix!

PR #587, opened 2026-05-28
Update `ftditool` and use new ELF loader in FPGA runner

ziuziakowska:
Update to latest version of `ftditool` (version 0.4.0). This introduces a new `bootstrap-elf` command, which comes with substantial speedups for loading sparse ELF files (e.g OpenSBI + a payload).

The FPGA runner would previously secretly depend on both the ELF file (for the entrypoint) and an `objcopy`'d binary with the same name and `.bin` suffix. With this command only the ELF file is required, which makes this more flexible.

See also: https://github.com/lowRISC/ftditool/pull/12, https://github.com/lowRISC/ftditool/pull/15

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -49,16 +49,16 @@
         ]
       },
       "locked": {
-        "lastModified": 1779281183,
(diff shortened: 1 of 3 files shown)


machshev (approved):
Thanks @ziuziakowska

engdoreis (approved):

PR #593, opened 2026-05-29
[hw] Rework reset tree for non-debug-module reset

### LINKED ISSUE #571: [design] Reset tree rework to allow NDM reset
In OpenTitan there is an LC reset and a SYS reset. The LC reset, resets everything besides the debug module and some of the PLIC, while the SYS reset, resets the debug domain.

At the moment in Mocha everything is run off of the SYS reset, so we should introduce an LC reset to allow the non-debug module reset to drive something.

### LINKED ISSUE #592: [design] Check clock manager clock and reset input
Check if the input clocks and resets to the clock manager are correctly connected.

raylau1:
This PR adds a debug domain reset under the sys reset tree, and moves existing main and IO clock resets under the LC reset tree.

General improvements to reset connections are also added.

Closes #571 
Closes https://github.com/lowRISC/mocha/issues/592

### DIFF
--- a/hw/top_chip/rtl/top_chip_system.sv
+++ b/hw/top_chip/rtl/top_chip_system.sv
@@ -1010,7 +1010,7 @@ module top_chip_system #(
     clkmgr_clocks.clk_main_hint | clkmgr_clocks.clk_io_peri |
     (|clkmgr_cg_en) |
     (|rstmgr_resets.rst_por_n) | (|rstmgr_resets.rst_spi_device_n) | (|rstmgr_resets.rst_spi_host_n) | (|rstmgr_resets.rst_i2c_n) |
-    (|rstmgr_rst_en);
+    (|rstmgr_resets.rst_debug_n) | (|rstmgr_rst_en);
 
   // Combine response and request between crossbar and atomics wrapper.
   AXI_BUS #(
--- a/hw/top_chip/rtl/top_chip_system.sv
+++ b/hw/top_chip/rtl/top_chip_system.sv
@@ -913,16 +913,20 @@ module top_chip_system #(
     .tl_o        (tl_clkmgr_d2h),
     .scanmode_i  (prim_mubi_pkg::MuBi4False),
 
-    // Clock and reset connections.
+    // Primary module control clocks and resets.
     .clk_i            (clkmgr_clocks.clk_io_powerup),
-    .clk_main_i       (clk_i),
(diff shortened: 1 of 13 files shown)


hw/top_chip/rtl/top_chip_system.sv
```diff
@@ -1009,7 +1009,7 @@ module top_chip_system #(
   assign unused_manager_output =
     clkmgr_clocks.clk_main_hint | clkmgr_clocks.clk_io_peri |
     (|clkmgr_cg_en) |
-    (|rstmgr_resets.rst_por_n) | (|rstmgr_resets.rst_spi_device_n) | (|rstmgr_resets.rst_spi_host_n) | (|rstmgr_resets.rst_i2c_n) |
```
marnovandermaas, line 1012: This change is premature as we haven't integrated this yet.

hw/top_chip/rtl/top_chip_system.sv
```diff
@@ -913,16 +913,21 @@ module top_chip_system #(
     .tl_o        (tl_clkmgr_d2h),
     .scanmode_i  (prim_mubi_pkg::MuBi4False),
 
-    // Clock and reset connections.
+    // TODO: Check if the following clock and reset connections are correct
```
marnovandermaas: @marnovandermaas figure this out.

marnovandermaas (commented):
Initial review, which I will also fix.

hw/top_chip/rtl/top_chip_system.sv
marnovandermaas: In the first commit I now only add the debug as ignored output. The other ignores are removed in the commit connecting up the right resets.

hw/top_chip/rtl/top_chip_system.sv
marnovandermaas: I think I've figured this out now.

marnovandermaas:
I've added an AON domain for the IO clock and also fixed the cascade assertions that were making the block-level smoke test fail. This'll need some more testing before it can be merged.

marnovandermaas:
After my changes the block-level DV also passes.
image

marnovandermaas:
And then top-level results are here, which has failures for ROM control and SPI device, which are also failing in our current nightly.
image

elliotb-lowrisc (approved):

PR #595, opened 2026-06-02
chore: migrate to NixOS 26.05 with fixes

machshev:
This flake follows the lowrisc-nix nixpkgs which has now been updated to nixos 26.05, which doesn't support Python 3.10

- [x] Update to Python 3.12
- [x] Updates the flake lock to pull in the latest lowrisc-nix packages
- [x] nix fmt . - looks like there is an existing formatting "issue"
- [x] Update update to use the a pinned version of verilator (5.040) which is the same version used before the update.

**Update:** _this PR has been updated to use a pinned version of verilator rather than trying to update the wavers to get it working with the new 5.048 version packaged with NixOS 26.05._


machshev:
I'm just looking into why the the verilator build failed

hw/top_chip/lint/top_chip_system.vlt
```diff
@@ -7,7 +7,7 @@
 `verilator_config
 
 // Initialising all RVFI signals to zero is OK in this case.
-lint_off -rule WIDTHCONCAT -file "*/cva6_cheri/core/cva6_rvfi_probes.sv" -match "*More than a 8k bit replication is probably wrong: 8475"
+lint_off -rule WIDTHCONCAT -file "*/cva6_cheri/core/cva6_rvfi_probes.sv" -match "*replication-limit 8192 is suspect: 8475"
```
marnovandermaas: Maybe we can keep both of these to keep the old Verilator versions working as well?

marnovandermaas (commented):
One change that would be good.

elliotb-lowrisc:
I think we need to pick a Verilator version and stick to it to avoid mismatch issues like this. I would suggest v5.040, as this is what we seemed to be used before this change.

hw/top_chip/lint/top_chip_system.vlt
machshev: Good idea, thanks!

machshev:
> I think we need to pick a Verilator version and stick to it to avoid mismatch issues like this. I would suggest v5.040, as this is what we seemed to be used before this change.

That's not so easy with nix, it's possible but we would have to use an overlay which means we don't get the benefit of the nix cache and everyone would have to rebuilt verilator locally.

```
verilator = pkgs.verilator.overrideAttrs (old: rec {
  version = "5.040";
  src = pkgs.fetchFromGitHub {
    owner = "verilator";
    repo = "verilator";
    rev = "v${version}";
    hash = "sha256-S+cDnKOTPjLw+sNmWL3+Ay6+UM8poMadkyPSGd3hgnc=";
  };
});
```

The upstream `lowrisc-nix` flake update pulled in NixOS 26.05 instead of 25.11, which means the next flake update would have made this change anyway as a side effect. The only thing I wanted to do in this PR was update to Python 3.12 as 3.10 is EOL in the next few months and removed from NixOS 26.05.

elliotb-lowrisc:
> That's not so easy with nix, it's possible but we would have to use an overlay which means we don't get the benefit of the nix cache and everyone would have to rebuilt verilator locally.

Is that what the OpenTitan nix shell does? It's on some ancient version I recall.

elliotb-lowrisc:
I'm trying to make sense of the SYNCASYNCNET waivers (I'm not worried about the other kinds) in case we do update Verilator, but I'm struggling to figure out what they are complaining about. Bear with me while I pull this PR so I can try and see the full warning message

machshev:
> > That's not so easy with nix, it's possible but we would have to use an overlay which means we don't get the benefit of the nix cache and everyone would have to rebuilt verilator locally.
> 
> Is that what the OpenTitan nix shell does? It's on some ancient version I recall.

Good point... it's using an old version we build in [lowrisc-nix](https://github.com/lowRISC/lowrisc-nix/blob/main/pkgs/verilator/default.nix). That would pull the version back to verilator `4.210`. Is that what we want to do here as well?

elliotb-lowrisc:
> Good point... it's using an old version we build in [lowrisc-nix](https://github.com/lowRISC/lowrisc-nix/blob/main/pkgs/verilator/default.nix). That would pull the version back to verilator `4.210`. Is that what we want to do here as well?

I think we'd need to change a lot more waivers (and possibly other things) if we were to go back to that old version

elliotb-lowrisc:
Aside: You'll need to remove `--trace-threads` from nightly.yml and dev_guide.md if we move to Verilator v5.048, as that option was deprecated in that version and causes an error (as I just found)

machshev:
PR to add a new verilator 5.040 package build to lowrisc-nix: https://github.com/lowRISC/lowrisc-nix/pull/186

elliotb-lowrisc:
May be irrelevant now but I'm still having trouble understanding the SYNCASYNCNET warnings. I think they are both probably fine, but the explanatory comment currently above them does not apply to them. I think they stem from some off the non-standard (but not necessarily bad) things done with reset signals in the OpenTitan reset manager code

machshev:
> May be irrelevant now but I'm still having trouble understanding the SYNCASYNCNET warnings. I think they are both probably fine, but the explanatory comment currently above them does not apply to them. I think they stem from some off the non-standard (but not necessarily bad) things done with reset signals in the OpenTitan reset manager code

Thanks for looking into this @elliotb-lowrisc I've reverted to Verilator 5.040 which is now available from lowrisc-nix. So I've removed the wavers to avoid the confusion.

elliotb-lowrisc (approved):
I'm no nix expert, but I approve of the intent

ziuziakowska (approved):
LGTM! Might be nice to add `nix fmt .` to CI in the future.

Issue #261, opened 2026-01-23
[SW] Implement DV specific registers to forward log and status


martin-velay:
How is it done in OpenTitan from the SW side (from what I know, it might be incomplete): 
- DV only specific registers as done on OpenTitan: https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rv_core_ibex/doc/registers.html#dv_sim_window
- Need to add things like:
  - https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/testing/test_framework/status.c
  - https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/arch/device_sim_dv.c

How is it handled from the HW perspective:
- Doc: https://opentitan.org/book/hw/dv/sv/sim_sram/index.html?highlight=intercept#simulation-sram
- Top TB code: https://github.com/lowRISC/opentitan/blob/master/hw/top_darjeeling/dv/tb/tb.sv#L203-L271
- RTL code: https://github.com/lowRISC/opentitan/blob/master/hw/top_darjeeling/ip_autogen/rv_core_ibex/rtl/rv_core_ibex.sv#L945-L990

PR #600, opened 2026-06-04
[sw,dv] Add SW-DV logging

martin-velay:
Related to this issue: https://github.com/lowRISC/mocha/issues/261

### DIFF
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
@@ -1,0 +1,58 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class top_chip_dv_dv_log_smoketest_vseq extends top_chip_dv_base_vseq;
+  `uvm_object_utils(top_chip_dv_dv_log_smoketest_vseq)
+
+  string exp_messages[] = '{
+    "SW-DV log smoketest starting...",
+    "a = 10, b = 20, c = 12",
+    "a + b + c = 42",
+    "Smoke test for the SW-DV log interface, completed successfully!"
+  };
+
+  typedef bit [1024*8-1:0] arg_t;
+
+  extern function new(string name = "");
+  extern task body();
+  // Per SV LRM 6.16, a string shorter than the packed type is left-padded with zeros: the last
+  // character sits at byte[0] (LSB) and the first at byte[N-1]. Collect the bytes upward from
+  // byte[0] until the null terminator, then reverse them with the streaming operator to restore
+  // left-to-right order.
+  extern function string arg_t_to_string(arg_t val);
+endclass : top_chip_dv_dv_log_smoketest_vseq
+
+
+function top_chip_dv_dv_log_smoketest_vseq::new(string name = "");
+  super.new(name);
+endfunction : new
+
+function string top_chip_dv_dv_log_smoketest_vseq::arg_t_to_string(arg_t val);
+  byte unsigned bytes[$];
+  for (int i = 0; i < ($bits(val) / 8); i++) begin
+    logic [7:0] c = val[i*8 +: 8];
+    if (c == 8'h00) break;
+    bytes.push_back(c);
+  end
+  return string'({<<8{bytes}});
+endfunction : arg_t_to_string
+
+task top_chip_dv_dv_log_smoketest_vseq::body();
+  if (cfg == null) begin
+    set_handles();
+  end
+
+  // Check each expected message inline as the event fires. This ensures we sample printed_log
+  // in the same delta as the event.
+  for (int i = 0; i < exp_messages.size(); i++) begin
+    string actual;
+    @(cfg.sw_logger_vif.printed_log_event);
+    actual = arg_t_to_string(arg_t'(cfg.sw_logger_vif.printed_log));
+    if (actual != exp_messages[i]) begin
+      `uvm_error(get_full_name(),
+                 $sformatf("Log message mismatch at index %0d. Expected '%0s' but saw '%0s'.",
+                           i, exp_messages[i], actual))
+    end
+  end
+endtask : body
--- a/hw/top_chip/dv/env/top_chip_dv_env_pkg.sv
+++ b/hw/top_chip/dv/env/top_chip_dv_env_pkg.sv
@@ -41,8 +41,18 @@ package top_chip_dv_env_pkg;
   // 50 MHz Peripheral clock
   parameter int unsigned PeriClkFreq = 50_000_000;
 
-  // SW-DV special locations for test status, logging, and platform identification.
-  // In simulation, sim_sram_axi_sink is connected to the dedicated sw_dv crossbar port.
+  // SW-DV special locations for test status, logging, and platform identification. In simulation,
+  // sim_sram_axi_sink is connected to the dedicated sw_dv crossbar port and each location holds a
+  // 32-bit value.
+  //
+  // The sw_test_status_if and sw_logger_if monitors only sample the low 32 bits of the 64-bit AXI
+  // write data (req.w.data[31:0], enforced by the tb DataUpperBitsZero_A assertion). A 32-bit store
+  // must land in that lane, so written locations (test status, log) sit at 8-byte aligned offsets;
+  // a store to offset 4 would land in the upper lane and be missed. HW_ID is read-only, so it can
+  // sit at offset 4 (the sink returns it on the read-data upper half resp.r.data[63:32]). The log
+  // takes the 0x8 slot rather than sharing with HW_ID so it has room to grow to 8 bytes for 64-bit
+  // values.
+
   parameter bit [31:0] SW_DV_START_ADDR       = 'h2002_0000;
   parameter bit [31:0] SW_DV_SIZE             = 'h0000_0100;        // 256 bytes reserved for SW DV
   parameter bit [31:0] SW_DV_TEST_STATUS_ADDR = SW_DV_START_ADDR + 'h00;
--- a/hw/vendor/lowrisc_ip/util/device_sw_utils/extract_sw_logs.py
+++ b/hw/vendor/lowrisc_ip/util/device_sw_utils/extract_sw_logs.py
@@ -173,6 +173,33 @@ def get_str_at_addr(str_addr, addr_strings):
     raise KeyError(f"string at addr {str_addr:x} not found")
 
 
+# CHERI __cap_relocs entry: five little-endian uint64 words.
+_CAP_RELOC_FMT = '<QQQQQ'
+_CAP_RELOC_SIZE = struct.calcsize(_CAP_RELOC_FMT)
+
+
+def get_cap_relocs(elf):
+    '''Build a {capability_location: target_addr} map from __cap_relocs.
+
+    On CHERI targets, pointer fields in initialised data (such as the file and
+    format pointers of the log_fields_t entries in .logs.fields) are stored as
+    capabilities whose address is materialised at runtime. The static ELF only
+    holds a placeholder, while the real target is recorded in __cap_relocs. Each
+    entry is five uint64 words: (location, object, offset, size, perms); the
+    resolved target address is object + offset. Returns an empty dict on
+    non-CHERI ELFs, where the section is absent or empty.'''
+    section = elf.get_section_by_name('__cap_relocs')
+    relocs = {}
+    if section is None:
+        return relocs
+    data = section.data()
+    for off in range(0, len(data), _CAP_RELOC_SIZE):
+        location, obj, offset, _size, _perms = struct.unpack(
+            _CAP_RELOC_FMT, data[off:off + _CAP_RELOC_SIZE])
+        relocs[location] = obj + offset
+    return relocs
+
+
 def extract_sw_logs(elf_file, logs_fields_section):
     '''This function extracts contents from the logs fields section, and the
     read only sections, processes them and generates a tuple of (results) -
@@ -222,17 +249,42 @@ def extract_sw_logs(elf_file, logs_fields_section):
                 logs_fields_section, elf_file))
             sys.exit(1)
 
-        header_size = 4
-        logs_offset, = struct.unpack('I', logs_data[0:header_size])
+        # Determine the log_fields_t binary layout, which depends on the target.
+        #  - rv32:  20-byte entries (all uint32) behind a 4-byte header.
+        #  - rv64:  32-byte entries with 8-byte pointer fields, 8-byte header.
+        #  - CHERI: file/format pointers are 16-byte capabilities, which forces
+        #           16-byte section alignment and 64-byte entries behind a
+        #           16-byte header. The capabilities hold placeholders in the
+        #           static image; their target addresses come from __cap_relocs
+        #           (see get_cap_relocs), keyed by the in-image slot location.
+        cap_relocs = get_cap_relocs(elf)
+        is_cheri = int(section.header['sh_addralign']) >= 16
+        if is_cheri:
+            sec_addr = int(section.header['sh_addr'])
+            header_size, entry_size = 16, 64
+            file_off, format_off = 0x10, 0x30
+            logs_offset = sec_addr
+        elif elf.elfclass == 64:
+            header_size, entry_size, entry_fmt = 8, 32, '<I4xQIIQ'
+            logs_offset, = struct.unpack('<Q', logs_data[0:header_size])
+        else:
+            header_size, entry_size, entry_fmt = 4, LOGS_FIELDS_SIZE, '<IIIII'
+            logs_offset, = struct.unpack('<I', logs_data[0:header_size])
 
         # Dump the logs with fields.
         result = ""
-        num_logs = (logs_size - header_size) // LOGS_FIELDS_SIZE
+        num_logs = (logs_size - header_size) // entry_size
         for i in range(num_logs):
-            start = header_size + i * LOGS_FIELDS_SIZE
-            end = start + LOGS_FIELDS_SIZE
-            severity, file_addr, line, nargs, format_addr = struct.unpack(
-                'IIIII', logs_data[start:end])
+            start = header_size + i * entry_size
+            if is_cheri:
+                severity, = struct.unpack('<I', logs_data[start:start + 4])
+                line, nargs = struct.unpack(
+                    '<II', logs_data[start + 0x20:start + 0x28])
+                file_addr = cap_relocs[sec_addr + start + file_off]
+                format_addr = cap_relocs[sec_addr + start + format_off]
+            else:
+                severity, file_addr, line, nargs, format_addr = struct.unpack(
+                    entry_fmt, logs_data[start:start + entry_size])
             result += "addr: {}\n".format(hex(logs_offset + start)[2:])
             result += "severity: {}\n".format(severity)
             result += "file: {}\n".format(
--- a/sw/device/lib/boot/mocha.ld
+++ b/sw/device/lib/boot/mocha.ld
@@ -30,6 +30,17 @@ SECTIONS {
         . = ALIGN(8);
     } > ram
 
+    /* DV simulation log database. Each call site puts a log_fields_t entry here.
+     * extract_sw_logs.py reads the ELF section to produce a .logs.txt database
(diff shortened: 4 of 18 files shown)


hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
```diff
@@ -0,0 +1,54 @@
...
+function top_chip_dv_dv_log_smoketest_vseq::new(string name = "");
+  super.new(name);
+endfunction : new
+
+function string top_chip_dv_dv_log_smoketest_vseq::arg_t_to_string(arg_t val);
+  string result = "";
+  for (int i = 0; i < ($bits(val) / 8); i++) begin
+    logic [7:0] c = val[i*8 +: 8];
+    if (c == 8'h00) break;
+    result = {$sformatf("%c", c), result};
```
rswarbrick: It might be nicer to make a queue of bytes in the loop and then use either $sformatf with %s or streaming operator to turn that into a string.

For example, couldn't you do something like this?
- Append `c` to a queue of bytes (`my_queue.push_back(c)`)
- Convert and reverse with `string'({<<8{my_queue}})`

hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
```diff
@@ -0,0 +1,54 @@
...
+  end
+  return result;
+endfunction : arg_t_to_string
+
+task top_chip_dv_dv_log_smoketest_vseq::body();
+  string actual;
```
rswarbrick: Minor nit: I'd move this variable inside the for loop to give it a shorter scope.

hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
```diff
@@ -0,0 +1,54 @@
...
+  if (cfg == null) begin
+    set_handles();
+  end
+
+  // Check each expected message inline as the event fires. This ensures we sample printed_log
+  // in the same delta as the event.
+  for (int i = 0; i < exp_messages.size(); i++) begin
+    @(cfg.sw_logger_vif.printed_log_event);
+    actual = arg_t_to_string(arg_t'(cfg.sw_logger_vif.printed_log));
+    `DV_CHECK_STREQ(actual, exp_messages[i], $sformatf("Log message [%0d] mismatch", i))
```
rswarbrick: I'm not a fan of these "helpful" macros from Sri. It might be cleaner to write it explicitly:
```systemverilog
    if (actual != exp_messages[i]) begin
      `uvm_error(get_full_name(),
                 $sformatf("Log message mismatch at index %0d. Expected '%0s' but saw '%0s'.",
                           i, exp_messages[i], actual))
    end
```

hw/top_chip/dv/env/top_chip_dv_env_pkg.sv
```diff
@@ -41,11 +41,13 @@ package top_chip_dv_env_pkg;
   // 50 MHz Peripheral clock
   parameter int unsigned PeriClkFreq = 50_000_000;
 
-  // SW DV special write locations for test status and logging will always fit in 32-bits
+  // SW DV special write locations for test status and logging will always fit in 32-bits.
+  // Addresses must be 8-byte aligned: AxiDataWidth=64, so a 32-bit store at offset 4
+  // within a bus word lands in req.w.data[63:32] (upper lane), not req.w.data[31:0].
```
rswarbrick: Is the point that keeping these addresses 8-byte aligned avoids the AXI agent or virtual sequence needing to shuffle things around when constructing wdata? It might be worth saying so explicitly here (I didn't understand at first!)

Or is the point that we want to use `SIZE=3` and need the accesses to be naturally aligned?

hw/vendor/lowrisc_ip/util/device_sw_utils/extract_sw_logs.py
```diff
@@ -222,17 +222,30 @@ def extract_sw_logs(elf_file, logs_fields_section):
                 logs_fields_section, elf_file))
             sys.exit(1)
 
-        header_size = 4
-        logs_offset, = struct.unpack('I', logs_data[0:header_size])
+        # rv64 ELF: struct layout is 32 bytes with 8-byte pointer fields and a
+        # QUAD section-address header.  rv32 ELF: 20 bytes, all uint32, LONG header.
```
rswarbrick: This comment should probably go after `if is_64bit:` to avoid it looking like it describes the other branch as well.

rswarbrick (commented):
A couple of nitty comments but this looks nice.

(And the review took me a while: I hadn't seen the clever variadic macro trick before!)

martin-velay:
I've just rebased and fix the merge conflicts. There was also an issue for CHERI tests. Now Verilator tests pass and the UVM tests too (new tests `dv_log_smoketest` and `dv_log_smoketest_cheri` in `test_framework` regression):
```
dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i test_framework
...
          [   legend    ]: [S: scheduled, Q: queued, R: running, P: passed, F: failed, K: killed, T: total]                                                                                               
00:01:07  [    build    ]: [S:         0, Q:      0, R:       0, P:      1, F:      0, K:      0, T:     1] 100%                                                                                          
00:01:43  [     run     ]: [S:         0, Q:      0, R:       0, P:      6, F:      0, K:      0, T:     6] 100%  
...

```

And the log gives:
```
UVM_INFO @            553397000: (sw_logger_if.sv:524) [dv_log_smoketest_vanilla_sram(sw/device/tests/test_framework/dv_log_smoketest.c:14)] SW-DV log smoketest starting...
UVM_INFO @            560717000: (sw_logger_if.sv:524) [dv_log_smoketest_vanilla_sram(sw/device/tests/test_framework/dv_log_smoketest.c:15)] a = 10, b = 20, c = 12
UVM_INFO @            565357000: (sw_logger_if.sv:524) [dv_log_smoketest_vanilla_sram(sw/device/tests/test_framework/dv_log_smoketest.c:16)] a + b + c = 42
UVM_INFO @            568337000: (sw_logger_if.sv:524) [dv_log_smoketest_vanilla_sram(sw/device/tests/test_framework/dv_log_smoketest.c:17)] Smoke test for the SW-DV log interface, completed successfully!
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
martin-velay: I'll do that

hw/top_chip/dv/env/top_chip_dv_env_pkg.sv
martin-velay: I'll clarify the comment

hw/vendor/lowrisc_ip/util/device_sw_utils/extract_sw_logs.py
martin-velay: Some rework were needed, so now it should be better

hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
```diff
@@ -0,0 +1,58 @@
...
+  if (cfg == null) begin
+    set_handles();
+  end
+
+  // Check each expected message inline as the event fires. This ensures we sample printed_log
+  // in the same delta as the event.
+  for (int i = 0; i < exp_messages.size(); i++) begin
+    string actual;
+    @(cfg.sw_logger_vif.printed_log_event);
+    actual = arg_t_to_string(arg_t'(cfg.sw_logger_vif.printed_log));
```
marnovandermaas, line 51: This requires the C smoke test to run, which is added in a later commit. We should put this commit later.

sw/device/lib/boot/mocha.ld
```diff
@@ -30,6 +30,17 @@ SECTIONS {
...
+     * extract_sw_logs.py reads the ELF section to produce a .logs.txt database
+     * consumed by sw_logger_if at simulation start. The first word is the section
+     * load address (header used by extract_sw_logs.py to map each entry to its
+     * runtime address). */
+    .logs.fields : {
+        QUAD(LOADADDR(.logs.fields));
+        KEEP(*(.logs.fields))
+        . = ALIGN(8);
+    } > ram
+
```
marnovandermaas, line 43: We don't want this in the Boot ROM, only for tests.

sw/device/lib/test_framework/dv_log.h
```diff
@@ -0,0 +1,54 @@
...
+            __attribute__((section(".logs.fields"), used)) = { \
+                .severity = (sev), \
+                .file = __FILE__, \
+                .line = (uint32_t)__LINE__, \
+                .nargs = DV_LOG_NARGS_SEQ_(x, ##__VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0), \
+                .format = (fmt), \
+            }; \
+        dv_log_write(&DV_LOG_UNIQUE_(dv_log_entry_), ##__VA_ARGS__); \
+    } while (0)
+
```
marnovandermaas, line 62: Is this macro magic really necessary?

hw/top_chip/dv/env/top_chip_dv_env_pkg.sv
rswarbrick: Thanks (and now I understand!)

sw/device/lib/test_framework/dv_log.h
marnovandermaas, line 1: @engdoreis what do you think of this? Are we reinventing the wheel here?

rswarbrick (approved):
This looks really good to me.

Thanks for the updates (and for teaching me about variadic macros in C...)

sw/device/lib/test_framework/dv_log.c
```diff
@@ -0,0 +1,21 @@
...
+
+void dv_log_write(const log_fields_t *log, ...)
+{
+    dv_window_t dv_window = mocha_system_dv_window();
+    DEV_WRITE(&dv_window->log, (uint32_t)(uintptr_t)log);
+
+    va_list args;
+    va_start(args, log);
+    for (uint32_t i = 0; i < log->nargs; i++) {
+        DEV_WRITE(&dv_window->log, va_arg(args, uint32_t));
```
marnovandermaas: Why is this 32-bit and not 64-bit?

sw/device/lib/test_framework/dv_log.c
```diff
@@ -0,0 +1,21 @@
...
+
+#include "test_framework/dv_log.h"
+#include "hal/mmio.h"
+#include "hal/mocha.h"
+#include <stdarg.h>
+
+void dv_log_write(const log_fields_t *log, ...)
+{
+    dv_window_t dv_window = mocha_system_dv_window();
+    DEV_WRITE(&dv_window->log, (uint32_t)(uintptr_t)log);
```
marnovandermaas: Pointers are 64-bit right?

util/build_sw_collateral_for_sim.py
```diff
@@ -26,6 +64,12 @@ def generate(args) -> None:
         cmd = [*install_cmd, target]
         subprocess.run(cmd, capture_output=False, check=True)
 
+        # Generate .logs.txt / .rodata.txt for the sw_logger_if DV monitor.
+        # The ELF is installed without an extension (CMake default).
+        elf_file = out_dir / target
+        if elf_file.exists():
```
marnovandermaas: If the elf file doesn't exist we have a bigger problem right?

util/build_sw_collateral_for_sim.py
```diff
@@ -5,8 +5,46 @@
...
+# Path to the OpenTitan-derived script that generates the .logs.txt and
+# .rodata.txt database files consumed by the DV sw_logger_if monitor.
+_EXTRACT_SW_LOGS = (
+    Path(__file__).parent.parent /
+    "hw" / "vendor" / "lowrisc_ip" / "util" / "device_sw_utils" /
+    "extract_sw_logs.py"
+)
+
+
+def _extract_logs(elf_file: Path, out_dir: Path) -> None:
```
marnovandermaas, line 22: We should only be doing this for tests not for all Verilator and FPGA software. Can you confirm this?

marnovandermaas (commented):
Some comments from my end.

hw/top_chip/dv/env/seq_lib/top_chip_dv_dv_log_smoketest_vseq.sv
martin-velay: So I should probably swap the commits and put the [sw] and [utils] commits 1st

sw/device/tests/test_framework/dv_log_smoketest.c
```diff
@@ -0,0 +1,20 @@
...
+#include "test_framework/dv_log.h"
+#include "hal/uart.h"
+
+bool test_main(uart_t console)
+{
+    (void)console;
+
+    DV_LOG_INFO("SW-DV log smoketest starting...");
+    DV_LOG_INFO("a = %d, b = %d, c = %d", 10, 20, 12);
+    DV_LOG_INFO("a + b + c = %d", 42);
```
engdoreis, line 16: do you want to also test %x, %c, %s?

sw/device/lib/test_framework/dv_log.h
```diff
@@ -0,0 +1,54 @@
...
+#include <stdint.h>
+
+// SW-to-DV logging: writes a log_fields_t entry address + arguments to
+// SW_DV_LOG_ADDR; sw_logger_if reconstructs the formatted string in simulation.
+
+// const char * for string fields: required for a valid C99 static initialiser
+// (pointer-to-integer casts are not constant expressions).  extract_sw_logs.py
+// detects the rv64 ELF class and uses '<I4xQIIQ' (32 B) instead of '<IIIII'.
+typedef struct {
+    uint32_t severity;
```
engdoreis: Can you create a enum for severity

sw/device/lib/test_framework/dv_log.h
```diff
@@ -0,0 +1,54 @@
...
+    do { \
+        static const log_fields_t DV_LOG_UNIQUE_(dv_log_entry_) \
+            __attribute__((section(".logs.fields"), used)) = { \
+                .severity = (sev), \
+                .file = __FILE__, \
+                .line = (uint32_t)__LINE__, \
+                .nargs = DV_LOG_NARGS_SEQ_(x, ##__VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0), \
+                .format = (fmt), \
+            }; \
+        dv_log_write(&DV_LOG_UNIQUE_(dv_log_entry_), ##__VA_ARGS__); \
```
engdoreis, line 60: Where is `dv_log_entry_` defined?

sw/device/lib/boot/mocha.ld
martin-velay: I don't know enough about the Boot ROM and the SW architecture, could you / @engdoreis guide me to move this in the correct location?

sw/device/lib/test_framework/dv_log.c
engdoreis: prefer the macro `VOLATILE_WRITE` instead

sw/device/tests/test_framework/dv_log_smoketest.c
engdoreis: Should also test priting a capability since you added support for it.

util/build_sw_collateral_for_sim.py
martin-velay: Good point. I have added `_has_log_entries` to filter out tests which doesn't require the log feature

util/build_sw_collateral_for_sim.py
martin-velay: Good point. This is also addressed with the new `_has_log_entries` check.

sw/device/lib/test_framework/dv_log.c
martin-velay: The sim SRAM window is only 32-bits and the `sw_logger_if` only samples 32-bits and the upper bits have an assertion in the `tb.sv` to ensure it's always the case

sw/device/lib/test_framework/dv_log.c
martin-velay: DV monitor consumes 32-bit beats, a 64 bits value would be truncated with today's implementation

sw/device/lib/test_framework/dv_log.h
engdoreis: The macro is neccessary, otherwise we would need to declare the struct with `__attribute__((section(".logs.fields"), used))` everywhere the log function is called.

sw/device/lib/boot/mocha.ld
engdoreis: This linker script is not used by the Boot ROM, so it's fine.

sw/device/lib/test_framework/dv_log.h
martin-velay: Done

sw/device/lib/test_framework/dv_log.h
martin-velay: I have added a comment above the defines to explain

sw/device/lib/test_framework/dv_log.c
martin-velay: Done, thanks

sw/device/tests/test_framework/dv_log_smoketest.c
martin-velay: Good idea on `%x/%c/%s `- I'll add a line exercising them to the smoketest.

On printing a capability: the log arg path only carries 32-bit values (`dv_log_write` reads each arg as `va_arg(uint32_t)`), so we can log a capability's address via %x/%p but not the full 128-bit capability. The CHERI support in this PR is about extracting the log database from a CHERI-built ELF and writing the log port safely (VOLATILE_WRITE), both already exercised end-to-end by the `dv_log_smoketest_cheri` variant. Did you mean logging an address, or were you thinking of the DB extraction that the `_cheri` test already covers? Full 128-bit capability logging would need a wider arg path, which I'd suggest as a follow-up.

martin-velay: @engdoreis, I did a test actually and it seems that the `%c` isn't cleanly supported by the upstream cleanup: it rewrites %c to %0c (the same way %d becomes %0d), and SV `$sformatf` rejects %0c (`*W,SYSFMW: Illegal format ignored`), so the log warns even though the char still prints. %x/%s are fine. I can either drop %c from the smoketest or fix cleanup_format to leave %c untouched (one-token change, and %0c is invalid SV anyway). Any preference?

engdoreis: For now let's just test the supported formats.

sw/device/lib/test_framework/dv_log.h
engdoreis: hmm, I understand it now, thanks!

util/build_sw_collateral_for_sim.py
engdoreis: @marnovandermaas I don't understand your point, this script is only run by dvsim.

engdoreis (approved):
LGTM, but let's wait for Marno's approval before merging.

sw/device/tests/test_framework/dv_log_smoketest.c
martin-velay: Shall I open an issue to add this later?

sw/device/lib/test_framework/dv_log.c
marnovandermaas: Ok, we don't currenlty use the top 32 bits of the address so I'm happy with this.

sw/device/tests/test_framework/dv_log_smoketest.c
marnovandermaas: I also think it's fine not to support %c.

marnovandermaas (approved):
Happy for you to merge when you think it is ready.

sw/device/lib/test_framework/dv_log.c
marnovandermaas: This is fine for now, maybe make an issue to track 64-bit support.

martin-velay:
Mentioned issues have been created and I ran again a UVM little regression and all looks good. Thank you all for you reviews!

PR #607, opened 2026-06-12
hw: correct PLIC context interrupt enable stride

ziuziakowska:
See https://github.com/lowRISC/opentitan/pull/30371.


marnovandermaas (approved):
Looks good to me, thanks for finding this issue!

PR #608, opened 2026-06-12
fix PLIC context interrupt mapping in Devicetree

ziuziakowska:
These were the wrong way around, it is Context 0 that delivers M-mode external interrupts and Context 1 that delivers S-mode external interrupts. Needed for correct Linux IRQ mappings.


marnovandermaas (approved):
This looks reasonable to me and only tested now that there is a supervisor mode.

PR #610, opened 2026-06-16
Update OpenTitan vendored files

martin-velay:
This is required as:
- Multiple DV fixes were made on OpenTitan based on observations on Mocha
- Some improvement have happened on the DV side
- Improve regression duration (reseed optimisation)


hw/vendor/lowrisc_ip/ip/kmac/data/kmac.hjson
marnovandermaas, line 1: I don't want to pull in design changes like these.

marnovandermaas (commented):
I don't mind pulling in DV changes but there are too many design changes in this import. If you really want to pull in these changes from upstream we may need to make a separate vendor for DV from design. Alternatively you can pull in the changes through patches.

marnovandermaas:
Marking this as draft because CI is failing.

martin-velay:
Not the right way anymore, as we don't want to pull DD changes

PR #617, opened 2026-06-17
fpga runner, nix, sw: Update to `ftditool` 0.5.0

ziuziakowska:
Update to `ftditool` version 0.5.0.

This version renames `bootstrap-elf` to `load-elf` and does not send a reset to the flash chip automatically - a new `flash-reset` command is provided to do so. This is to allow loading multiple binaries/ELFs before sending the flash reset, which we can use in CI for testing Linux with an initramfs.

The fpga_runner command line interface has been modified to take binaries and their load address with the `-f` flag, and ELF files with the `-e` flag. Usage sites in CMake have also been updated.

### DIFF
--- a/util/fpga_runner.py
+++ b/util/fpga_runner.py
@@ -78,14 +78,35 @@ async def bootstrap(uart: serial.Serial | None) -> None:
     await set_pin(0, True)
 
 
-async def load_fpga_binary(path: Path, uart: serial.Serial | None) -> None:
-    await bootstrap(uart)
-    command = ftditool_command("bootstrap-elf")
+async def load_elf(path: Path) -> None:
+    command = ftditool_command("load-elf")
     command.append(str(path))
 
     p = await asyncio.create_subprocess_exec(*command)
     if await p.wait() != 0:
-        error("ftditool bootstrap-elf", f"exited with non-zero exit code {p.returncode}")
+        error("ftditool load-elf", f"exited with non-zero exit code {p.returncode}")
+        sys.exit(1)
+
+
+async def load_binary(path: Path, address: int) -> None:
+    command = ftditool_command("load-file")
+    command.extend([str(path), "--addr", hex(address)])
+
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
+        error("ftditool load-file", f"exited with non-zero exit code {p.returncode}")
+        sys.exit(1)
+
+
+async def exit_bootstrap() -> None:
+    """
+    Exit SPI bootstrap mode on the FPGA by sending a flash reset command to the
+    emulated flash, causing the board to jump into uploaded code.
+    """
+    command = ftditool_command("flash-reset")
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
+        error("ftditool flash-reset", f"exited with non-zero exit code {p.returncode}")
         sys.exit(1)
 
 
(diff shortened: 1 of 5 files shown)


util/fpga_runner.py
```diff
@@ -78,14 +78,31 @@ async def bootstrap(uart: serial.Serial | None) -> None:
...
 
     p = await asyncio.create_subprocess_exec(*command)
     if await p.wait() != 0:
-        error("ftditool bootstrap-elf", f"exited with non-zero exit code {p.returncode}")
+        error("ftditool load-elf", f"exited with non-zero exit code {p.returncode}")
+        sys.exit(1)
+
+
+async def load_fpga_binary(path: Path, address: int) -> None:
+    command = ftditool_command("load-file")
```
marnovandermaas, line 92: Should we name this load binary instead?

marnovandermaas (commented):
Just a minor comment, probably good to get @engdoreis approval before merging.

util/fpga_runner.py
```diff
@@ -78,14 +78,31 @@ async def bootstrap(uart: serial.Serial | None) -> None:
...
+    command.extend([str(path), hex(address)])
+
+    p = await asyncio.create_subprocess_exec(*command)
+    if await p.wait() != 0:
+        error("ftditool load-file", f"exited with non-zero exit code {p.returncode}")
+        sys.exit(1)
+
+
+async def fpga_flash_reset() -> None:
```
engdoreis: Can we name this function `exit_bootstrap`? Although from the Flash protocol this is a reset, from the bootrom is a bootstrap exit.

engdoreis (approved):
LGTM, let's just wait for the PR in ftditool to be merged before this.

util/fpga_runner.py
```diff
@@ -19,29 +19,37 @@
...
 TEST_PATTERN = re.compile(r"TEST RESULT: (PASSED|FAILED)", re.IGNORECASE)
 
-RUNNER: str = Path(__file__).name
+RUNNER: str = Path(__file__).stem
 
 
-async def set_pin(pin: int, val: bool) -> None:
-    command = [
+def error(what: str, why: str) -> None:
+    print(f"[{RUNNER}]: {what}: {why}")
```
AlexJones0: Might be worth explicitly moving this now to `stderr`, unless there's a reason we want it to write to `stdout`?

util/fpga_runner.py
```diff
@@ -127,37 +161,70 @@ async def do_fpga_test(args) -> None:
...
     )
 
     run_parser = subparsers.add_parser("run", help=run_help, description=("run: " + run_help))
-    run_parser.add_argument("path", type=Path, help="Path to program ELF to load on the FPGA")
     run_parser.set_defaults(func=do_fpga_run)
 
+    for p in [test_parser, run_parser]:
+        p.add_argument("-e", action="append", metavar="elf", default=[], help="ELF file to load.")
+        p.add_argument(
+            "-f",
```
AlexJones0: Nit - I'd prefer that arguments have specific long-names that can be used (in this case `--elf` and `--bin`/`--binary` might be sensible?), as otherwise they end up propagating to single-letter names in the code. This should make the code a bit easier to read rather than referring to `args.e` and `args.f`

AlexJones0 (approved):
LGTM - just a couple of nits.

util/fpga_runner.py
ziuziakowska: Would be a good idea to have these, I think `argparse` also lets you choose what the destination attribute is named, so these could be stored in `args.elfs` and `args.bins` for example.

PR #615, opened 2026-06-17
[dv] Enable DRAM backdoor for the UVM TB

martin-velay:
This PR addresses issue https://github.com/lowRISC/mocha/issues/514

Issue #514, opened 2026-05-01
[DV] Create backdoor load for DRAM


engdoreis:
After the backdoor is created, the SRAM tests should be removed and the DRAM should be used instead.

- [x]  Create DRAM backdoor
- [x] Delete the `sram` from the target name in [hjson](https://github.com/lowRISC/mocha/blob/28542ccbe7cfca8f1550bb9eafe5fdf339a0560e/hw/top_chip/dv/top_chip_sim_cfg.hjson#L50)
- [x]  Delete the sram binaries from [cmake](https://github.com/lowRISC/mocha/blob/28542ccbe7cfca8f1550bb9eafe5fdf339a0560e/sw/cmake/tests.cmake#L108)
- [x] Clean up documentation (remove any explanation of the _sram binaries)

engdoreis:
@marnovandermaas for visibility

martin-velay:
This issue has been addressed by this PR https://github.com/lowRISC/mocha/pull/615

PR #625, opened 2026-06-21
[sw] BootROM default is SPI bootstrap

### LINKED ISSUE #624: Boot ROM blinks at start up
When an FPGA image is bare it, the Boot ROM should go into SPI bootstrap mode and flash LEDs.

marnovandermaas:
This is a better default because on FPGA the LEDs will show activity when the bitstream is first flashed and both DRAM and SRAM are disabled.

Closes: https://github.com/lowRISC/mocha/issues/624

### DIFF
--- a/sw/device/bootrom/bootrom.c
+++ b/sw/device/bootrom/bootrom.c
@@ -15,7 +15,7 @@
 #include <stdint.h>
 
 #define MAJOR "00"
-#define MINOR "01"
+#define MINOR "02"
 #define PATCH "00"
 
 const uintptr_t boot_slots[] = { 0x10000000, 0x80000000 };
@@ -45,6 +45,7 @@ static void clear_slots();
 // TODO: Add support to cheri mode
 int boot_main(void)
 {
+    uint32_t boot_addr;
     struct boot_context boot_ctx = (struct boot_context){
         .console = mocha_system_uart(),
         .timer = mocha_system_timer(),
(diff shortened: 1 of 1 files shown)


sw/device/bootrom/bootrom.c
```diff
@@ -56,14 +57,13 @@ int boot_main(void)
 
     timer_init(boot_ctx.timer);
     timer_enable_write(boot_ctx.timer, true);
-    if (bootstrap_requested(&boot_ctx)) {
+    if (bootstrap_requested(&boot_ctx) || !get_boot_addr(&boot_addr)) {
```
engdoreis, line 60: This will break the opensbi flow in simulation because it doesn't have the magic number. We need to figure out a better solution.

engdoreis: what if we flash the LEDs inside the exception handler?

engdoreis:
I'm sorry for accidentally closing the PR.

sw/device/bootrom/bootrom.c
marnovandermaas: Yes, I agree that is a better solution!

marnovandermaas:
Converted to draft as is breaks the OpenSBI flow on Verilator. @engdoreis is working on a better solution in the meantime.

ziuziakowska:
I'm not sure about this. The rationale for boot ROM defaulting to jumping straight to DRAM was to not need any host-side software to bypass Bootstrap mode in Verilator simulation - where binaries are loaded by backdoor instead. We also cannot test the presence of a magic number because the magic number is optional in case the next boot address should be different - e.g jumping into SRAM instead of the default DRAM, and this boot magic mechanism and SRAM tests is planned to be removed when https://github.com/lowRISC/mocha/pull/615 is merged (issue https://github.com/lowRISC/mocha/issues/514).

marnovandermaas:
Closing in favour of github.com/lowRISC/mocha/pull/630

PR #630, opened 2026-06-22
[sw,bootrom] Add LED animation on exception

### LINKED ISSUE #624: Boot ROM blinks at start up

engdoreis:
This will show that the bitstream is live when a program is not loaded to DRAM.

Supersedes https://github.com/lowRISC/mocha/pull/625
Closes: https://github.com/lowRISC/mocha/issues/624

### DIFF
--- a/sw/device/bootrom/bootrom.c
+++ b/sw/device/bootrom/bootrom.c
@@ -15,7 +15,7 @@
 #include <stdint.h>
 
 #define MAJOR "00"
-#define MINOR "01"
+#define MINOR "02"
 #define PATCH "00"
 
 const uintptr_t boot_slots[] = { 0x10000000, 0x80000000 };
@@ -56,6 +56,8 @@ int boot_main(void)
 
     timer_init(boot_ctx.timer);
     timer_enable_write(boot_ctx.timer, true);
+    led_init(boot_ctx.gpio);
+
     if (bootstrap_requested(&boot_ctx)) {
         uprintf(boot_ctx.console, "Entering SPI bootstrap\n");
         clear_slots(); // Cleaning slots from previeous boot.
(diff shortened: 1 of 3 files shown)


sw/device/lib/hal/gpio.c
marnovandermaas: Can you also fix this? I think this should be to make it work for the upper mask:
```
uint32_t mask = 1u << ((pin % 16) & 0xFu);
```

marnovandermaas (approved):
I haven't yet tested this but the code looks good so I'll approve to unblock. Also I think I've found a small bug in nearby code, maybe we can fix that in this PR as well?

sw/device/bootrom/bootrom.c
```diff
@@ -223,9 +223,34 @@ bool bootstrap_requested(struct boot_context *ctx)
...
+
+    while (true) {
+        gpio_write(gpio, pattern);
+        pattern ^= 0xff;
+        uint64_t timeout = timer_value_read_us(timer) + led_animation_period_us / 2;
+        while (timer_value_read_us(timer) < timeout) {
+        };
+    };
+}
+// TODO: Catch exceptions properly.
```
ziuziakowska: This comment can be removed I think.

sw/device/bootrom/bootrom.c
```diff
@@ -15,7 +15,7 @@
 #include <stdint.h>
 
 #define MAJOR "00"
-#define MINOR "01"
+#define MINOR "02"
 #define PATCH "00"
```
ziuziakowska, line 19: I think this could be updated in a different PR, I've also noticed that it doesn't follow the versioning used in https://github.com/lowRISC/mocha/#release-timeline, maybe best to just have it be one string like `"v0.2"`? I doubt we will be updating the versions frequently enough to need so much granularity.

marnovandermaas: I think it's nice to have this done in this PR so we can make the release.

marnovandermaas (approved):
This looks good to me and works on my board.

PR #631, opened 2026-06-22
Add nix shell wrapper to fpga_runner

engdoreis:
This will be mainly be used by users testing the release artefacts.
## Usage
```sh
nix run .#bitstream-load <path/to/bitstream>
```
```sh
nix run .#fpga-runner -- test -e build/sw/device/tests/uart_smoketest_cheri
```

### DIFF
--- a/nix/fpga.nix
+++ b/nix/fpga.nix
@@ -5,6 +5,7 @@
   pkgs,
   pythonEnv,
   llvm,
+  ftditool,
 }: let
   bitstream_path = "build/lowrisc_mocha_chip_mocha_genesys2_0/synth-vivado";
   bootrom_path = "build/sw/device/bootrom";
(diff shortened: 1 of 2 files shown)


nix/fpga.nix
```diff
@@ -51,7 +51,8 @@ in {
     name = "bitstream-load";
     runtimeInputs = [pkgs.openfpgaloader];
     text = ''
-      openFPGALoader -b genesys2 ${bitstream_path}/lowrisc_mocha_chip_mocha_genesys2_0.bit
+      bitstream=''${1:-${bitstream_path}/lowrisc_mocha_chip_mocha_genesys2_0.bit}
+      openFPGALoader -b genesys2 "$bitstream"
```
ziuziakowska, line 55: This looks like it would always append `/lowrisc_mocha_chip_mocha_genesys2_0.bit` to the end so it's expecting a directory, wouldn't it be better to check if `$1` is not empty and use that as the whole file path, otherwise default to the default bitstream path?

ziuziakowska: Oops I've misread the parentheses 😅 This indeed does do that!

ziuziakowska (approved):

marnovandermaas (approved):
Works for me.

PR #632, opened 2026-06-22
Update release workflow

### LINKED ISSUE #627: Update Release action with Linux
We need to update the release action: https://github.com/lowRISC/mocha/blob/main/.github/workflows/release.yml
With all the artefacts necessary to run:
```sh
util/fpga_runner.py run -e fw_payload.elf -f Image 0x90000000 -f rootfs_uboot_image 0xa0000000
```

ziuziakowska:
Closes https://github.com/lowRISC/mocha/issues/627.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,8 +76,7 @@ jobs:
   software-release:
     runs-on: ["self-hosted", "nixos", "X64"]
     env:
-      SOURCE: build/sw/device/examples
-      ARTEFACT: examples.tar.gz
+      ARTEFACT: release_software.tar.gz
 
     defaults:
       run:
@@ -93,11 +92,10 @@ jobs:
       - name: Generate artefacts
         run: |
           cmake -B build/sw -S sw
-
-          for t in hello_world hello_world_verilator; do
-            cmake --build build/sw -j $(nproc) --target $t
-            cmake --install build/sw/ --prefix release --component $t
-          done
+          cmake --build build/sw -j$(nproc)
+          cmake --install build/sw --prefix release/examples --component examples
+          cmake --install build/sw --prefix release --component bootrom
+          cmake --install build/sw --prefix release --component boot
 
           tar -czvf ${{ env.ARTEFACT }} release
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ permissions:
 on:
   release:
     types: [created]
+  workflow_dispatch:
 
 jobs:
   # This job will only succeed if a bitstream was published by a previous CI run to make sure it
--- a/sw/cmake/tests.cmake
+++ b/sw/cmake/tests.cmake
@@ -54,13 +54,11 @@ function(mocha_add_verilator_test)
         COMMAND ${PROJECT_SOURCE_DIR}/../util/verilator_runner.sh -r $<TARGET_FILE:${arg_ROM}>_scrambled.vmem -E ${arg_NAME}
     )
     set_tests_properties(${TEST} PROPERTIES TIMEOUT ${arg_TIMEOUT})
-
 endfunction()
 
 function(mocha_add_fpga_test)
     set(one_value_args NAME TIMEOUT)
     cmake_parse_arguments(arg "" "${one_value_args}" "" ${ARGN})
-
 
     if(NOT arg_TIMEOUT)
         set(arg_TIMEOUT 15)  # default
@@ -101,7 +99,6 @@ function(mocha_add_test)
         target_link_options(${NAME} PUBLIC
           "-Tmocha_dram.ld" "-L${LDS_DIR}"
         )
-
         # create artefacts
         mocha_add_executable_artefacts(NAME ${NAME})
 
--- a/sw/device/bootrom/CMakeLists.txt
+++ b/sw/device/bootrom/CMakeLists.txt
@@ -32,6 +32,6 @@ add_custom_command(
     VERBATIM
 )
 
-install(TARGETS ${NAME})
+install(TARGETS ${NAME} DESTINATION . COMPONENT ${NAME})
 install(FILES "$<TARGET_FILE:${NAME}>_scrambled.vmem" DESTINATION . COMPONENT ${NAME})
 install(FILES "$<TARGET_FILE:${NAME}>.bin" DESTINATION . COMPONENT ${NAME})
--- a/sw/device/tests/opensbi/CMakeLists.txt
+++ b/sw/device/tests/opensbi/CMakeLists.txt
@@ -13,4 +13,10 @@ target_compile_options(${NAME} PUBLIC ${VANILLA_FLAGS})
 # using just a few large pages. This is a requirement for Linux too,
 # see https://docs.kernel.org/arch/riscv/boot.html#kernel-location.
 target_link_options(${NAME} PUBLIC "-Wl,--Ttext=0x80200000")
-mocha_add_executable_artefacts(NAME ${NAME})
+# Objcopy to a binary, to be used as the OpenSBI payload.
+add_custom_command(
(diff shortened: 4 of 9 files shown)


engdoreis:
Fix: https://github.com/lowRISC/mocha/issues/627

sw/device/bootrom/CMakeLists.txt
```diff
@@ -32,6 +32,6 @@ add_custom_command(
     VERBATIM
 )
 
-install(TARGETS ${NAME})
-install(FILES "$<TARGET_FILE:${NAME}>_scrambled.vmem" DESTINATION . COMPONENT ${NAME})
-install(FILES "$<TARGET_FILE:${NAME}>.bin" DESTINATION . COMPONENT ${NAME})
+install(TARGETS ${NAME} DESTINATION boot COMPONENT boot)
```
engdoreis: I didn't test, but it affects the UVM tests.

ziuziakowska: I think this does modify the path, do you happen to know where this would need to be adjusted to work for UVM?

engdoreis: It would good to run the UVM test to make sure, but if it changes the path, then all the tests in hw/top_earlgrey/dv/chip_sim_cfg.hjson will need to be updated.

ziuziakowska: I will check this, but I think the destination on `install` only affects the subfolder it is placed in when `--install` is used.

ziuziakowska: Looks like the build paths are still the same.

ziuziakowska: Ah wait, this will affect the CMake commands in `build_sw_collateral_for_sim.py`.

.github/workflows/release.yml
marnovandermaas: This should probably be renamed to something like: `release_sw.tar.gz`

marnovandermaas (commented):
I've tested the build and that seems to work.

.github/workflows/release.yml
marnovandermaas: I've resolved this myself but feel free to rename.

marnovandermaas (dismissed):
Here's an example of UVM test failure:
```
UVM_FATAL @                    0: (mem_bkdr_util.sv:658) [mem_bkdr_util[ChipMemSRAM]] file /home/mvdmaas/mocha/scratch/update-release-workflow/top_chip_system.sim.xcelium/0.mailbox_smoke/latest/mailbox_smoketest_vanilla_sram.vmem could not be opened for r mode
```

.github/workflows/release.yml
```diff
@@ -92,11 +92,10 @@ jobs:
...
         run: |
           cmake -B build/sw -S sw
-
-          for t in hello_world hello_world_verilator; do
-            cmake --build build/sw -j $(nproc) --target $t
-            cmake --install build/sw/ --prefix release --component $t
-          done
+          cmake --build build/sw -j$(nproc)
+          cmake --install build/sw --prefix release --component examples
+          cmake --install build/sw --prefix release --component tests
```
engdoreis: I would remove the tests from the release, what do you think @marnovandermaas ?

marnovandermaas: Yes, I was going to suggest that too.

sw/CMakeLists.txt
```diff
@@ -17,6 +17,13 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 include(ExternalProject)
 
+# Custom target that represents boot related artefacts.
+# E.g Boot ROM, firmware (OpenSBI), Linux, root filesystem.
+add_custom_target(boot)
+
+# Custom target that represents all test binaries.
+add_custom_target(tests)
```
marnovandermaas: I thought we no longer had this?

sw/cmake/tests.cmake
```diff
@@ -101,7 +108,7 @@ function(mocha_add_test)
         target_link_options(${NAME} PUBLIC
           "-Tmocha_dram.ld" "-L${LDS_DIR}"
         )
-
+        add_dependencies(tests ${NAME})
```
marnovandermaas: Is this still necessary?

sw/cmake/tests.cmake
```diff
@@ -115,6 +122,7 @@ function(mocha_add_test)
             target_link_options(${NAME}_sram PUBLIC
               "-Tmocha_sram.ld" "-L${LDS_DIR}"
             )
+            add_dependencies(tests ${NAME}_sram)
```
marnovandermaas: Is this still necessary?

sw/device/tests/opensbi/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ target_compile_options(${NAME} PUBLIC ${VANILLA_FLAGS})
 # using just a few large pages. This is a requirement for Linux too,
 # see https://docs.kernel.org/arch/riscv/boot.html#kernel-location.
 target_link_options(${NAME} PUBLIC "-Wl,--Ttext=0x80200000")
-mocha_add_executable_artefacts(NAME ${NAME})
+mocha_add_executable_artefacts(NAME ${NAME} TYPE tests)
```
marnovandermaas: Is this now the only one that use this type tests? Maybe we can just remove this and remove all the other dependencies?

sw/cmake/tests.cmake
```diff
@@ -28,9 +29,17 @@ function(mocha_add_executable_artefacts)
         VERBATIM
     )
 
-    install(TARGETS ${arg_NAME} DESTINATION . COMPONENT ${arg_NAME})
-    install(FILES "$<TARGET_FILE:${arg_NAME}>.vmem" DESTINATION . COMPONENT ${arg_NAME})
-    install(FILES "$<TARGET_FILE:${arg_NAME}>.bin" DESTINATION . COMPONENT ${arg_NAME})
+    if(arg_TYPE)
```
engdoreis: I think this adds unnecessary complexity. We should either leave it as is and use --prefix to set the target path, or update the DESTINATION here along with the UVM tests.

engdoreis: changing the file https://github.com/lowRISC/mocha/blob/830bf10f404c69f0388bd421a10d99d90c98c106/hw/top_chip/dv/top_chip_sim_cfg.hjson#L52.
from:
```
+ChipMemROM_image_file={run_dir}/bootrom_scrambled.vmem
```
to: 
```
+ChipMemROM_image_file={run_dir}/boot/bootrom_scrambled.vmem
```
Should work

sw/device/tests/opensbi/CMakeLists.txt
ziuziakowska: I think for this test we actually only need the `.bin` because it will get wrapped in OpenSBI anyway.

sw/cmake/tests.cmake
ziuziakowska: I've reverted all the destinations to `.` for now

marnovandermaas:
DV passes now (with some expected failures):
```
          [   legend    ]: [S: scheduled, Q: queued, R: running, P: passed, F: failed, K: killed, T: total]                                                                                                           
00:01:16  [    build    ]: [S:        00, Q:     00, R:      00, P:     01, F:     00, K:     00, T:    01] 100%                                                                                                      
00:02:51  [     run     ]: [S:        00, Q:     00, R:      00, P:     28, F:     04, K:     00, T:    32] 100%
```

sw/device/bootrom/CMakeLists.txt
marnovandermaas: This is now resolved I believe.

marnovandermaas (approved):
I think this now works, so I will merge to try to make a release.

engdoreis (approved):

PR #633, opened 2026-06-22
[top-i2c,dv] Added chip level I2C Device TX_RX test

### LINKED ISSUE #144: I2C DV - Top-level smoke tests

### LINKED ISSUE #509: I2C DV -- Device TX-RX test
Write a `vseq` that will act as a `host` to send `write` and `read` transfer. At the end, contains some self checking mechanism to compare the data written is equal to data received

KinzaQamar:
Added a host Vseq and device SW. The host will initiate read / write transfer. The device SW waits for the transfers to finish and compare the data bytes in each transfer at the end


sw/device/tests/i2c/device_tx_rx_test.c
```diff
@@ -0,0 +1,189 @@
...
+static bool check_num_bytes()
+{
+    return ((byte_count > 0) && (byte_count <= TX_FIFO_DEPTH));
+}
+
+// Wait for the read transfer to complete by checking interrupt state and status register fields.
+static bool wait_read_finish(i2c_t i2c)
+{
+    while (true) {
+        if (VOLATILE_READ(i2c->intr_state) & i2c_intr_cmd_complete) {
```
KinzaQamar: wrap the READ and every `VOLATILE_READ` in a HAL function

sw/device/lib/hal/i2c.c
```diff
@@ -217,3 +217,42 @@ uint8_t i2c_rdata_byte(i2c_t i2c)
...
+    VOLATILE_WRITE(i2c->target_id, target_id);
+}
+
+void i2c_write_tx_data(i2c_t i2c, uint8_t data)
+{
+    i2c_txdata txdata = { .txdata = data };
+    VOLATILE_WRITE(i2c->txdata, txdata);
+}
+
+void i2c_clear_cmd_complete(i2c_t i2c)
```
engdoreis, line 235: I suggest making a function to clear interrupts that take the interrupt as argument like uart:
https://github.com/lowRISC/mocha/blob/3b3b9e2d41ab1fbf57833e2c4fd7ab86caffdcc7/sw/device/lib/hal/uart.c#L36

engdoreis (approved):
I left a suggestion, otherwise it LGTM.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_device_tx_rx_vseq.sv
```diff
@@ -0,0 +1,207 @@
...
+                                       sw_sys_clk_period_ns[0]);
+  stop_setup_cycles  = round_up_divide({sw_setup_stop_time_ns[1], sw_setup_stop_time_ns[0]},
+                                       sw_sys_clk_period_ns[0]);
+  stop_hold_cycles   = round_up_divide({sw_hold_stop_time_ns[1], sw_hold_stop_time_ns[0]},
+                                       sw_sys_clk_period_ns[0]);
+  rise_cycles        = round_up_divide({sw_rise_time_ns[1], sw_rise_time_ns[0]},
+                                       sw_sys_clk_period_ns[0]);
+  fall_cycles        = round_up_divide({sw_fall_time_ns[1], sw_fall_time_ns[0]},
+                                       sw_sys_clk_period_ns[0]);
+endtask
```
martin-velay: Nit: most of your new methods are long enough to justify the fact to add the label, I feel it hard to read without it.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
```diff
@@ -37,23 +69,42 @@ function int unsigned top_chip_dv_i2c_tx_rx_vseq::round_up_divide(int unsigned a
 endfunction
 
 function void top_chip_dv_i2c_tx_rx_vseq::configure_agent_timing();
```
martin-velay, line 137: I think these changes also have an impact on `top_chip_dv_i2c_host_tx_rx_vseq`. Isn't it an issue?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_device_tx_rx_vseq.sv
```diff
@@ -0,0 +1,207 @@
...
+
+task top_chip_dv_i2c_device_tx_rx_vseq::i2c_xfer();
+  i2c_item xfer = i2c_item::type_id::create("xfer");
+  i2c_target_base_seq host_seq = i2c_target_base_seq::type_id::create("host_seq");
+
+  create_and_drive_i2c_xfer(xfer, READ, BusOpRead, host_seq);
+
+  // De-allocate the xfer item in order to send a new transfer
+  xfer = null;
```
martin-velay: This is dead code and can be dropped

sw/device/tests/i2c/device_tx_rx_test.c
```diff
@@ -0,0 +1,155 @@
...
+// The constants assigned below are the spec minimums for standard mode speed except
+// hold_data_time_ns which should be at least one according to OpenTitan's programming guide.
+const uint16_t scl_low_time_ns = 4700;
+const uint16_t scl_high_time_ns = 4000;
+const uint16_t hold_data_time_ns = 1;
+const uint16_t setup_data_time_ns = 250;
+const uint16_t setup_start_time_ns = 4700;
+const uint16_t hold_start_time_ns = 4000;
+const uint16_t setup_stop_time_ns = 4000;
+const uint16_t hold_stop_time_ns = 4700;
```
martin-velay: Nit: `hold_stop_time_ns` is actually bus-free time (tBUF) per the I2C spec, not a hold time on the stop condition — same concept as `bus_free_time_cycles` in `i2c_timing_params_t` (`i2c.h`). Same naming carries through to `sw_hold_stop_time_ns`/`stop_hold_cycles` in `top_chip_dv_i2c_tx_rx_vseq.sv` and the backdoor-read in `top_chip_dv_i2c_device_tx_rx_vseq.sv`. Could you either rename to `bus_free_time_ns`/`bus_free_cycles` for consistency, or add a short comment noting it maps to tBUF?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
KinzaQamar: I don't think that `host_tx_rx_test` gets affected. I've took these calculations from `i2c_base_vseq` function `get_timing_values()`. They look sensible to reason.

KinzaQamar: Oh, sorry. I think I didn't answer this question well. These parameters are used to add delays which should depend on the speed of the I2C bus and not on host / device part. So, whatever the speed is just calculate the timing parameter the same way

sw/device/tests/i2c/device_tx_rx_test.c
KinzaQamar: Thanks Martin. Your consistency comments always helps. You are right here. I've also noticed the in `i2c.h`, the field of the `struct` should name as `bus_free_cycles` but I'll do this in a separate PR

sw/device/lib/hal/i2c.c
KinzaQamar: Do you mean something like that:
```
void i2c_clear_intr_state(i2c_t i2c, uint32_t state) 
{
    VOLATILE_WRITE(i2c->intr_state, state);
}
```

engdoreis: Yes, but 'state' would be an enum with the interrupts. It should also accept more then one interrupt per call.

hw/top_chip/dv/top_chip_sim_cfg.hjson
```diff
@@ -180,6 +180,13 @@
       run_opts: ["+ChipMemSRAM_image_file={run_dir}/i2c_device_tx_rx_test_vanilla_sram.vmem",
                  "+ChipMemROM_image_file={run_dir}/bootrom_scrambled.vmem"]
     }
+    {
+      name: i2c_device_tx_rx_cheri
+      uvm_test_seq: top_chip_dv_i2c_device_tx_rx_vseq
+      sw_images: ["i2c_device_tx_rx_test_vanilla_sram:5" "bootrom:5"]
```
martin-velay:
```suggestion
      sw_images: ["i2c_device_tx_rx_test_cheri_sram:5" "bootrom:5"]
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
martin-velay: Sorry, I was unclear. The issue I see is that you've created the `dut_init` task for `top_chip_dv_i2c_device_tx_rx_vseq` to read some SW symbols and then use them in this `configure_agent_timing` function, all fine here. But in your implementation of `top_chip_dv_i2c_host_tx_rx_vseq::dut_init()` should also backdoor read these missing SW symbols.

This also needs to be added in `host_tx_rx_test.c` as done in `device_tx_rx_test.c` or in a common shared file?

Something like:
```c
const uint16_t scl_high_time_ns = 4000;
const uint16_t setup_data_time_ns = 250;
const uint16_t setup_start_time_ns = 4700;
const uint16_t hold_start_time_ns = 4000;
const uint16_t setup_stop_time_ns = 4000;
const uint16_t bus_free_time_ns = 4700;
const uint16_t rise_time_ns = I2C_RISE_NS;
const uint16_t fall_time_ns = I2C_FALL_NS;
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_device_tx_rx_vseq.sv
```diff
@@ -0,0 +1,207 @@
...
+  // Manage the I2C transfer
+  extern local task i2c_xfer();
+  extern task body();
+endclass : top_chip_dv_i2c_device_tx_rx_vseq
+
+function top_chip_dv_i2c_device_tx_rx_vseq::new(string name = "");
+  super.new(name);
+endfunction
+
+function top_chip_dv_i2c_device_tx_rx_vseq::calc_timing_params_in_cycles();
```
martin-velay: I think this function should be moved in the `top_chip_dv_i2c_tx_rx_vseq` as I believe `top_chip_dv_i2c_host_tx_rx_vseq` also need to have the same thing?

KinzaQamar: Yes, thanks for noticing it

hw/top_chip/dv/top_chip_sim_cfg.hjson
KinzaQamar: Oops, thanks for noticing it.

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
KinzaQamar: No, they are not useful for the `host` test. For host test, the TB is the device responsible for sending responses (rdata, ack, and nack). This is the function [i2c_driver.sv#L50](https://github.com/lowRISC/opentitan/blob/215aac82638ef76bbe81cf05e6f482762aa05fe8/hw/dv/sv/i2c_agent/i2c_driver.sv#L50) called by the driver where only `device_send_bit(_)` is used deep down in `i2c_if` that majorly uses `tClockPulse`, `tClockLow`, and `tHoldBit`. other parameters like `tSdaUnstable` and `tTimeOut` are not neccessary for smoke test

sw/device/lib/hal/i2c.c
KinzaQamar: For this test, I only wanted to clear `cmd_complete` field of `intr_state` register. I am not checking / enabling interrupts for smoke tests. But your suggestion is good and I'll do it as part of a separate PR. Thanks!

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
KinzaQamar: I've added comments here: [changes#diff-e7f07ab57afcc3afbc109eaf390fe9db01a5e5372a8e20b41ccb00e04c9ef14eR37-R51](https://github.com/lowRISC/mocha/pull/633/changes#diff-e7f07ab57afcc3afbc109eaf390fe9db01a5e5372a8e20b41ccb00e04c9ef14eR37-R51) explaining the we oinly need subset of timing parameters for host test

martin-velay: Traced through `calc_timing_params_in_cycles()` for the host path and I don't think "the remaining parameters default to 0" holds up. `round_up_divide(a, b)` is `((a - 1) / b) + 1`, with `a/b` as `int unsigned`. For every parameter `host_tx_rx` doesn't backdoor-read (`scl_high`, `setup_data`, `setup_start`, `hold_start`, `setup_stop`, `bus_free`, `rise`, `fall`), the backing `sw_*_time_ns` array is never written, so `a = 0` when round_up_divide runs. `0 - 1` in unsigned 32-bit arithmetic wraps to `4294967295`, not -1, so the result isn't 0 — it's `floor(4294967295 / clk_period) + 1` truncated into a 16-bit field. For a 5ns clock period that's 858993460, which truncates to 13108 cycles, not 0. So 8 of the 10 `*_cycles` fields end up with a clock-period-dependent garbage value, and nearly every field `configure_agent_timing()` computes (`tSetupBit`, `tClockPulse`, `tClockLow`, `tSetupStop`, `tHoldStop`) mixes in at least one of them.

I'd rather we set these explicitly for the host test the same way `device_tx_rx_test.c` does, rather than leaning on `round_up_divide` to return 0 for an input it was never written to handle — that keeps both vseqs symmetric and doesn't depend on an accident of integer wraparound working out. Could you add the same 8 consts to `host_tx_rx_test.c` and the corresponding backdoor reads to `top_chip_dv_i2c_host_tx_rx_vseq::dut_init()`, same as the device vseq?

KinzaQamar: Oh yes, thanks for figuring this out. My intention is just to add stuff that is required for the vseq / test. I can add all the timing parameters in `host_tx_rx_test.x` but it is not going to be consumed by the `i2c_driver` when it is driving the `device` interface. I can go ahead and revert it back to manually calculating `scl_low_cycles` and `scl_hold_cycles`. WDYT?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_device_tx_rx_vseq.sv
```diff
@@ -0,0 +1,183 @@
+// Copyright lowRISC contributors (COSMIC project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class top_chip_dv_i2c_device_tx_rx_vseq extends top_chip_dv_i2c_tx_rx_vseq;
+  `uvm_object_utils(top_chip_dv_i2c_device_tx_rx_vseq)
+
+  // TODO: Remove once #600 is merged. This is maintained by SW to make this Vseq in sync
```
martin-velay, line 8: This PR https://github.com/lowRISC/mocha/pull/600 has been merged now. Could you rebase and remove this workaround?

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_device_tx_rx_vseq.sv
```diff
@@ -0,0 +1,183 @@
...
+  cfg.m_i2c_agent_cfg.if_mode = Host;
+  super.body();
+  `DV_WAIT(cfg.sw_test_status_vif.sw_test_status == SwTestStatusInTest);
+
+  configure_agent_timing();
+  print_i2c_timing_cfg();
+
+  // Wait until SW is done writing to the TX FIFO
+  //
+  // TODO: Remove when #600 is merged
```
martin-velay, line 145: Same here https://github.com/lowRISC/mocha/pull/600 is now merged

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
martin-velay: Reverting fixes the underflow, but `tClockPulse`/`tSetupBit` still end up 0 either way, and `tClockPulse` is one of the three you called essential. Rather than reason about it further, could you run `i2c_host_tx_rx` and paste the `print_i2c_timing_cfg()` output here? If it passes cleanly with `tClockPulse`/`tSetupBit` at 0, I'll drop this. If not, let's just add the explicit consts to `host_tx_rx_test.c` instead.

KinzaQamar: >  but tClockPulse/tSetupBit still end up 0 either way, and tClockPulse is one of the three you called essential. 

Oh yes, you are right. Sorry, I am a bit slow today. I'll re-think about it again. Thanks for your careful analysis

KinzaQamar: I've added a shared header file `i2c_test.h` and put the computations of timing parameters in `top_chip_dv_tx_rx_vseq`

KinzaQamar:
Both host / device cheri / no-cheri tests are passing:
```
dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_device_tx_rx -t xcelium --reseed=10
dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_device_tx_rx_cheri -t xcelium --reseed=10
dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_host_tx_rx -t xcelium --reseed=10
dvsim hw/top_chip/dv/top_chip_sim_cfg.hjson -i i2c_host_tx_rx_cheri -t xcelium --reseed=10
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_tx_rx_vseq.sv
martin-velay: Thanks

martin-velay (approved):
I think we're good here now. Thanks a lot for this good contribution Kinza!

marnovandermaas (approved):
Thank you Kinza!

PR #634, opened 2026-06-22
[top-i2c,dv] Made host_tx_rx_test randomized

KinzaQamar:
Now it can read / write multiple bytes on any 7-bit address

### DIFF
--- a/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
+++ b/hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
@@ -11,10 +11,20 @@
 class top_chip_dv_i2c_host_tx_rx_vseq extends top_chip_dv_i2c_tx_rx_vseq;
   `uvm_object_utils(top_chip_dv_i2c_host_tx_rx_vseq)
 
+  // Declare the device_addr as a byte size array because sw_symbol_backdoor_overwrite() expects the
+  // data that is going to overwrite the SW symbol to be an array
+  local rand bit [7:0] device_addr[1];
+
+  extern constraint device_addr_c;
+
   extern function new(string name="");
-  extern task body();
   extern virtual task dut_init(string reset_kind = "HARD");
+  extern task body();
 endclass : top_chip_dv_i2c_host_tx_rx_vseq
+
+constraint top_chip_dv_i2c_host_tx_rx_vseq::device_addr_c {
+  device_addr[0] inside {[8'h00 : 8'h7F]};
+}
 
 function top_chip_dv_i2c_host_tx_rx_vseq::new(string name = "");
   super.new(name);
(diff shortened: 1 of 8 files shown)


hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -11,9 +11,13 @@
 class top_chip_dv_i2c_host_tx_rx_vseq extends top_chip_dv_i2c_tx_rx_vseq;
   `uvm_object_utils(top_chip_dv_i2c_host_tx_rx_vseq)
 
+  // Declare the device_addr as a byte size array because sw_symbol_backdoor_overwrite() expect the
```
elliotb-lowrisc: "expect" -> "expects"

elliotb-lowrisc (approved):
Seems good to me, assuming that it has been tested. Just one minor typo

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -11,9 +11,13 @@
 class top_chip_dv_i2c_host_tx_rx_vseq extends top_chip_dv_i2c_tx_rx_vseq;
   `uvm_object_utils(top_chip_dv_i2c_host_tx_rx_vseq)
 
+  // Declare the device_addr as a byte size array because sw_symbol_backdoor_overwrite() expects the
+  // data that is going to overwrite the SW symbol to be an array
+  local rand bit [7:0] device_addr[1];
```
martin-velay, line 16: I think something can go wrong as the address here is 8 bits but the i2c expects 7 bits. Probably better to add a constraint to this variable?

```sv
constraint top_chip_dv_i2c_host_tx_rx_vseq::device_addr_c {
  device_addr[0] inside {[8'h00 : 8'h7F]};
}
```

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
```diff
@@ -11,9 +11,13 @@
...
+
   extern function new(string name="");
-  extern task body();
   extern virtual task dut_init(string reset_kind = "HARD");
+  extern task body();
```
martin-velay, line 22: Nit: this change isn't required and it was better before since new and body methods are part of the UVM

KinzaQamar: I wanted to list the declarations aligned with the definitions. We have new(), dut_init, and body()

martin-velay: I still have a pref to keep the standard methods grouped and the class specific together. Not a blocking thing though

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: That's not going to happen due to this line [i2c.c#L143](https://github.com/lowRISC/mocha/blob/ca85787f9bcde25ebe9e8cef5106136977055aab/sw/device/lib/hal/i2c.c#L143). But you are right to explicitly constraint it to avoid confusion

hw/top_chip/dv/env/seq_lib/top_chip_dv_i2c_host_tx_rx_vseq.sv
KinzaQamar: You are right. Though, I prefer to put new() first and body() last because body() contains couple of other function calls that needs to get declared before body() as per my preference. I also remember that verissimo complains about the OOO listing thing in a different world and that has became my habit :)

martin-velay (approved):
Thanks Kinza. It looks good.
I also ran the simulations for the related tests:
```
|  Stage  |  Name  | Tests                |  Max Job Runtime  |  Simulated Time  |  Passing  |  Total  |  Pass Rate  |
|:-------:|:------:|:---------------------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
|         |        | i2c_host_tx_rx       |      24.000s      |    1676.118us    |     1     |    1    |  100.00 %   |
|         |        | i2c_host_tx_rx_cheri |     148.000s      |   13226.515us    |     1     |    1    |  100.00 %   |
```

PR #638, opened 2026-06-24
Update URLs to lowRISC repos

ziuziakowska:


marnovandermaas (commented):
As discussed elsewhere, let's also pin the commit hash in this PR. That would be a great improvement.

engdoreis (approved):

marnovandermaas (approved):
Thanks for adding the tags, when we need updates to these depending repositories, we should create a new tag (or point to a specific development commit). This is so that changes there require a review here as well.

PR #647, opened 2026-06-26
[fpga-loader] Skip erase on loading binaries

engdoreis:
Speed up binaries load from 1.8Mibps to 2.1Mibps.


raylau1 (approved):
For me the speed up is from 2.6Mibps to 2.8Mibps

ziuziakowska (approved):

PR #649, opened 2026-07-06
[fpga] Fix ROM corruption on async reset assertion

### LINKED ISSUE #639: FPGA top unresponsive after repeated external reset
The FPGA chip top can become unresponsive after repeated resets. This also causes the intermittent FPGA test timeouts in CI.

Steps to reproduce:
1. Load bitstream
1. Open UART terminal or look at LEDs
1. Rapidly press and release the reset button many times
1. The chip top eventually becomes unresponsive to all inputs (including external reset) unless the bitstream is reloaded

raylau1:
The asynchronous Mocha reset assertion on FPGA can cause timing violations for BRAM control/address input signals, and corrupt the content of the ROM.

This PR adds a synchroniser to ensure both assertion and deassertion of the Mocha top reset on FPGA are synchronous to the clock, which prevents this issue.

Closes #639.

### DIFF
--- a/hw/top_chip/rtl/chip_mocha_genesys2.sv
+++ b/hw/top_chip/rtl/chip_mocha_genesys2.sv
@@ -112,6 +112,7 @@ module chip_mocha_genesys2 #(
   logic mig_rst_n_sync_cfg;      // MIG system reset, async assertion, deassertion sync to clk_cfg
   logic mig_axi_rst_n_sync_200m; // MIG AXI reset, async assertion, deassertion sync to clk_200m
   logic rst_n_sync_50m;          // Mocha top reset, async assertion, deassertion sync to clk_50m
+  logic rst_n_sync_assert_50m;   // Mocha top synchronous reset, assertion and deassertion sync to clk_50m
   logic eth_rst_n_sync_125m;     // Ethernet MAC reset, async assertion, deassertion sync to clk_125m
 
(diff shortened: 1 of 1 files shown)


hw/top_chip/rtl/chip_mocha_genesys2.sv
marnovandermaas, line 207: Would this have the same effect?

```suggestion
.rst_ni (1'b1),
.d_i    (ext_rst_ni & ftdi_rst_ni & fpga_rst_n_sync_cfg),
```

marnovandermaas (commented):
Great find! I was wondering about one alteration.

hw/top_chip/rtl/chip_mocha_genesys2.sv
raylau1: No. Since the external reset also resets the PLL, there is a delay between reset deassertion and clocks starting to toggle. We need the external reset on the `rst_ni` pin to make sure the active value 0 is latched into the DFFs before the clocks start toggling.

marnovandermaas: Ok, thanks!

marnovandermaas (approved):
This change looks good to me. I re-ran the CI three times to make sure the FPGA test runs reliably now, and it does. So I think this is good to merge.

PR #650, opened 2026-07-06
[dv][kmac] Integrate KMAC Block-level DV from OpenTitan

### LINKED ISSUE #430: KMAC DV - Block-level env to be imported from OT

tchilikov-semify:
KMAC sanity regression can be ran with:
```
dvsim hw/vendor/lowrisc_ip/ip/kmac/dv/kmac_unmasked_sim_cfg.hjson -i nightly -r 1 --tool xcelium
```

### DIFF
--- a/hw/vendor/lowrisc_ip/ip/kmac/dv/env/kmac_env_cov.sv
+++ b/hw/vendor/lowrisc_ip/ip/kmac/dv/env/kmac_env_cov.sv
@@ -125,7 +125,7 @@ endgroup
 class app_cg_wrap;
   // done signal is sent while Keccak rounds are running
   covergroup app_cg(string name) with function sample(bit single_beat,
-                                                      bit [keymgr_pkg::KmacDataIfWidth/8-1:0] strb,
+                                                      bit [kmac_pkg::MsgWidth/8-1:0] strb,
                                                       bit err,
                                                       bit is_done,
                                                       bit in_keccak);
@@ -177,11 +177,11 @@ class app_cg_wrap;
 
   function new(string name = "app_cg");
     app_cg = new(name);
-    app_cfg_reg_cg = new(name);
+    app_cfg_reg_cg = new({name, "_cfg_reg"});
   endfunction
 
   function void app_sample(bit single_beat,
-                           bit [keymgr_pkg::KmacDataIfWidth/8-1:0] strb,
+                           bit [kmac_pkg::MsgWidth/8-1:0] strb,
                            bit err,
                            bit is_done,
                            bit in_keccak);
--- a/hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_common_vseq.sv
+++ b/hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_common_vseq.sv
@@ -18,10 +18,12 @@ class kmac_common_vseq extends kmac_base_vseq;
     if (common_seq_type inside {"shadow_reg_errors", "shadow_reg_errors_with_csr_rw"}) begin
       csr_excl_item csr_excl = ral.get_excl_item();
       // Shadow storage fatal error might cause req to drop without ack.
+`ifdef KMAC_MASKING
       $assertoff(0,
       "tb.dut.gen_entropy.u_prim_sync_reqack_data.u_prim_sync_reqack.SyncReqAckHoldReq");
       $assertoff(0,
       "tb.dut.gen_entropy.u_prim_sync_reqack_data.u_prim_sync_reqack.SyncReqAckAckNeedsReq");
+`endif
       $assertoff(0, "tb.edn_if[0].ReqHighUntilAck_A");
       $assertoff(0, "tb.edn_if[0].AckAssertedOnlyWhenReqAsserted_A");
     end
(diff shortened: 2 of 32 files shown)


tchilikov-semify:
Implements #430

hw/vendor/lowrisc_ip/ip/kmac/dv/env/kmac_env_cov.sv
```diff
@@ -177,11 +177,11 @@ class app_cg_wrap;
 
   function new(string name = "app_cg");
     app_cg = new(name);
-    app_cfg_reg_cg = new(name);
+    app_cfg_reg_cg = new({name, "_cfg_reg"});
```
martin-velay, line 180: Why this change? Is it to avoid having 2 times the same name? If yes, that's probably something we should fix upstream too?

hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_common_vseq.sv
```diff
@@ -18,10 +18,12 @@ class kmac_common_vseq extends kmac_base_vseq;
     if (common_seq_type inside {"shadow_reg_errors", "shadow_reg_errors_with_csr_rw"}) begin
       csr_excl_item csr_excl = ral.get_excl_item();
       // Shadow storage fatal error might cause req to drop without ack.
+`ifdef KMAC_MASKING
```
martin-velay, line 21: We should avoid the ifdef as much as possible (at least on the DV side). And I think we already have the material to do so in the `cfg` class. Could you try this instead and remove the `+define+KMAC_MASKING`?

```sv
if (cfg.enable_masking) begin
  $assertoff(0, "tb.dut.gen_entropy...SyncReqAckHoldReq");
  $assertoff(0, "tb.dut.gen_entropy...SyncReqAckAckNeedsReq");
end
```

hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_lc_escalation_vseq.sv
```diff
@@ -8,8 +8,10 @@ class kmac_lc_escalation_vseq extends kmac_app_vseq;
   `uvm_object_new
 
   virtual function void disable_asserts();
+`ifdef KMAC_MASKING
```
martin-velay, line 11: Same here

martin-velay (commented):
Thanks for this PR Christian. I have some comments, but you've done a good job!

hw/vendor/lowrisc_ip/ip/kmac/dv/env/kmac_env_cov.sv
tchilikov-semify: That's right - the issue here is that theres duplicate naming for different coverage properties. In the upstream where vcs was used this wasn't an issue it seems, but xcelium throws an error. this change makes it so the different properties have unique names. Youre right that this should probably be changes upstream for compatibility across simulators.

hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_common_vseq.sv
tchilikov-semify: Good catch, I didn't notice that this already exists. Will update it to use the cfg class instead of a macro

tchilikov-semify: I re-looked into this - I tried changing the guard to the `cfg` guard, but unfortunately this doesnt work. The reason I added the `ifdef` here was because these assertions should be "disabled" when the kmac is in unmasked mode, since the hierarchical paths are non-existent. There isnt such a guard upstream since its unneeded as VCS handles this gracefully at runtime by ignoring it (it seems). Xcelium however checks for hierarchical paths at elaboration time, and explicitly errors out. the `cfg` option doesn't work, since its a run-time guard.

martin-velay: Oh...
Could you try with a wildcard, maybe it's treated in a different way by Xcelium:

```sv
if (cfg.enable_masking) begin
  $assertoff(0, "tb.dut.*.SyncReqAckHoldReq");
  $assertoff(0, "tb.dut.*.SyncReqAckAckNeedsReq");
end
```

tchilikov-semify: Unfortunately the wildcard also doesnt work... Seems Xcelium treats it as a literal path and it doesnt support wildcards in these strings.

martin-velay: That's sad... So I guess there is no other choice than what you've done then. Thanks for having try :)

marnovandermaas (approved):
Thanks for going through this. This looks good to me, but please do resolve Martin's comments and make sure the regression runs before merging.

hw/vendor/lowrisc_ip/ip/kmac/dv/env/seq_lib/kmac_common_vseq.sv
tchilikov-semify: I think no other obvious choice... would gladly try an alternative. I agree `ifdef`s are ugly

hw/vendor/lowrisc_ip/ip/kmac/dv/env/kmac_env_cov.sv
martin-velay: OK, I'll make this change upstream then. Thanks

martin-velay:
I am currently running a regression to make sure results are aligned with OpenTitan. I'll post them later, please do not merge yet

martin-velay (approved):
All looks good, thanks Christian. I have also run a full regression and compared results against OpenTitan's.

PR #651, opened 2026-07-07
feat: modify the existing default devshell to use edaShell

machshev:
Experimental new EDA devshell.

- [x] Migrate the default devshell over to `lowrisc-nix.lib.mkEdaShell` to add support for commercial EDA tools versions are pulled from `tool_data.json`
- [x]  Removed unnecessary `v` prefix from `Vivado` version
- [x] Removed the `.envrc` (direnv) config which does `use flake` - unfortunately the new EdaShell uses FHS bubblewrap to support the EDA tools on NixOS, it is not source-able. Which means it doesn't work with direnv. This will mean we will have to manually `nix develop .`

**update:** depends on https://github.com/lowRISC/mocha/pull/702

### DIFF
--- a/ci/run.py
+++ b/ci/run.py
@@ -1,0 +1,361 @@
+#!/usr/bin/env python
+# Copyright lowRISC contributors (COSMIC project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""CI job runner: what each CI job does, so a job and a local run of it are the
+same command in the same shell.
+
+    nix run .#ci -- <job>...             run these jobs, in order
+    nix run .#ci -- all                  every job that needs no FPGA board
+    nix run .#ci -- --step <cmd> [args]  run a single step
+    nix run .#ci -- --list               list the jobs and their steps
+    nix run .#ci -- --dry-run <job>...   print the steps, run nothing
+    nix run .#ci -- --verbose ...        with --dry-run, ask each of our steps
+                                         what it would run; otherwise echo the
+                                         commands as they run
+
+From inside `nix develop` the same thing spelled ci/run.py <job>.
+
+A step is a command: a tool the devshell provides, one of the repository's
+util/ scripts, or -- where a step needs arguments worked out, or is more than
+one command -- a script in ci/scripts. Steps run from the repository root and
+each is runnable on its own.
+
+A step the rest of its job depends on, a configure or a build, is Required():
+if it fails the run stops there. Any other step records its failure and the job
+carries on, so one run reports every problem it can find.
+
+A job can also name another job, Job("name"), to run its steps at that point:
+the sequences are written once and composed rather than restated. A step
+reached twice over runs once, the first time it is reached.
+
+Parallelism comes from $MOCHA_JOBS (default: every core), so
+`MOCHA_JOBS=8 nix run .#ci -- verilator-test` leaves a machine usable.
+"""
+
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+from typing import NamedTuple, Union
+
+import lib
+
+
+class Step(NamedTuple):
+    """A command to run, and whether the rest of its job depends on it."""
+
+    argv: tuple
+    required: bool = False
+
+
+class Job(NamedTuple):
+    """Another job's steps, run at this point in the list."""
+
+    name: str
+
+
+# Reads as a kind of step rather than a function, hence the capital.
+def Required(*argv) -> Step:
+    return Step(tuple(argv), required=True)
+
+
+def step(*argv) -> Step:
+    return Step(tuple(argv))
+
+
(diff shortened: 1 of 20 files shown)


marnovandermaas:
It looks like the bubble wrap is causing an error. Any idea how to fix this:
```
bwrap: No permissions to create a new namespace, likely because the kernel does not allow non-privileged user namespaces. On e.g. debian this can be enabled with 'sysctl kernel.unprivileged_userns_clone=1'.
Error: Process completed with exit code 1.
```

machshev:
> It looks like the bubble wrap is causing an error. Any idea how to fix this:
> 
> ```
> bwrap: No permissions to create a new namespace, likely because the kernel does not allow non-privileged user namespaces. On e.g. debian this can be enabled with 'sysctl kernel.unprivileged_userns_clone=1'.
> Error: Process completed with exit code 1.
> ```

This looks like the Ubuntu 24 AppArmor issue we have with Nix in general.

@jwnrt suggested this:

```bash
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
sudo sysctl -w kernel.apparmor_restrict_unprivileged_unconfined=0
```

machshev:
Looks like the way we are now calling the devshell in CI makes the github action runner think the job has frozen or something like that and kills the process with SIGKILL. It doesn't seem to be consistent though, sometimes the job passes and other times it's killed.

Trying a wrapper similar to OpenTitan lint flow to see if that helps stability?

machshev:
> Looks like the way we are now calling the devshell in CI makes the github action runner think the job has frozen or something like that and kills the process with SIGKILL. It doesn't seem to be consistent though, sometimes the job passes and other times it's killed.
> 
> Trying a wrapper similar to OpenTitan lint flow to see if that helps stability?

This seems to have resolved the issue and also means we can more easily run CI locally before PR... like OT.

marnovandermaas (commented):
Thank you for making these changes. From what I can tell the first commit looks good. One part of the second commit I am not sure is an improvement. You have pulled out all the commands from the workflow yml files into CI bash script directory. However, I believe the resulting lib.sh and run.sh are less readable than the previous ci.yml. Personally, I think we should either go down the route of individual bash scripts per step or to create a Python script like with artefacts.py.

machshev:
> Thank you for making these changes. From what I can tell the first commit looks good. One part of the second commit I am not sure is an improvement. You have pulled out all the commands from the workflow yml files into CI bash script directory. However, I believe the resulting lib.sh and run.sh are less readable than the previous ci.yml. Personally, I think we should either go down the route of individual bash scripts per step or to create a Python script like with artefacts.py.

Yes I agree with the readability comment. The main reason for the second commit in this PR is because the EDA devshell is FHS devshell to provide support for commercial EDA tools that are not supported directly with nix.

The extra level of abstraction also works around an issue with GHA runner that seems to think the devshell job has finished before it has and kills it with SIGKILL.

ci/run.sh
engdoreis, line 1: This feels too much bash, is there any argument for not using python?

machshev: It's a bit more bash than I'm generally comfortable with. But yes the benefit of a Bash only solution is not having to rely on python interpreter at runtime whereas Bash is ubiquitous. Although it kind of feels on the edge.

Is it a good enough argument? Not sure... suspect this would be a lot more readable as python and if we stuck to Python standard library then it should be standalone.

UPDATE: actually Python is already part of the devshell, so that does reduce the argument against.

Do you want me to migrate it to Python?

engdoreis: Since we will running this with nix, and as you mentioned Python is already in the devshell then you advise to migrate it to python.

ci/run.py
```diff
@@ -0,0 +1,361 @@
...
+def ci_script(name: str, *args) -> tuple:
+    """One of our own steps, by its job-style name."""
+    return (f"ci/scripts/{name.replace('-', '_')}.py", *args)
+
+
+# ---------------------------------------------------------------------------
+# The jobs. One per CI job, named after it; the workflows call these by name.
+# ---------------------------------------------------------------------------
+
+JOBS: "dict[str, list[Union[Step, Job]]]" = {
```
engdoreis: I would prefer a json or toml file describing the jobs and steps.

engdoreis: Or if it's more convinient to describe it in python, then move this to another module/file that would only contain the JOBS definition.

ci/run.sh
machshev: Done

ziuziakowska:
The GDB script fix LGTM, good to have that be more deterministic.

machshev:
> The GDB script fix LGTM, good to have that be more deterministic.

Thanks for review this... good to know it makes sense! I'm not that familiar with gdb... pdb on the other hand ;)

ci/run.py
machshev: Done... it's now a TOML file

engdoreis (approved):
Thanks for the changes.

PR #664, opened 2026-07-10
[doc] KMAC removed from memory map

### LINKED ISSUE #609: Remove KMAC from architecture memory map
We're now not going to connect the KMAC registers to any memory bus as KMAC will only be used by the Scrambled ROM checker directly, so we should probably remove KMAC from the memory map table in the architecture documentation

marnovandermaas:
The KMAC block in Mocha is only used to verify the contents of the Boot ROM. There is not software interface for this, so we should remove it from the memory map.

Closes: https://github.com/lowRISC/mocha/issues/609


raylau1 (approved):
Thanks, LGTM.

PR #667, opened 2026-07-10
[dv,sram] add axi_sram verification plan and tag architecture spec

tchilikov-semify:
This PR adds a verification plan for the axi_sram, and makes some minor updates to the specification. It also tags the specification and adds a markdown memory map table to it.


doc/ref/arch.md
```diff
@@ -1,76 +1,76 @@
...
-If a portion of the 128-bit aligned region is written it must clear the tag for the whole region including when a partial write strobe is used.
-
-Reads that only read part of a 64-bit value are allowed from valid capability regions, but these should have their tag cleared.
-Burst reads from the SRAM must have the appropriate CHERI tags set for each address, so a valid capability must have the user bits set for both of the 64-bit flits it is being sent back, and a mixture of capability and non-capability data is allowed in a burst.
-The SRAM is allowed to mark a capability as invalid by setting one or both of the `ruser` bits to zero, so the core must AND the two `ruser` values together to determine the validity of a capability.
-Tags should be stored in a separate block of memory from the data, this is to allow future optimisations where bulk-reads of tags are desired.
-
-The initial value of the SRAM including the tags is undefined at start-up.
+The static random-access memory (SRAM) in CHERI Mocha is mainly used as the stack and heap for the boot firmware that lives in the read-only memory (ROM).<!-- 21k32t -->
```
martin-velay: Exclude this sentence:

```suggestion
The static random-access memory (SRAM) in CHERI Mocha is mainly used as the stack and heap for the boot firmware that lives in the read-only memory (ROM).<!-- 21k32t_x -->
```

doc/ref/arch.md
```diff
@@ -1,76 +1,76 @@
...
+However, it should also be possible to execute from SRAM.<!-- lhjkel -->
+Once code starts executing from dynamic random-access memory (DRAM), we don't envision using SRAM anymore.<!-- a3l94p -->
```
martin-velay: Same:

```suggestion
Once code starts executing from dynamic random-access memory (DRAM), we don't envision using SRAM anymore.<!-- a3l94p_x -->
```

doc/ref/arch.md
```diff
@@ -1,76 +1,76 @@
...
+- Clock input<!-- 947rwh -->
+- Reset input<!-- rrni5j -->
+- AXI4 request input from the main SoC sub-system crossbar<!-- pgo845 -->
+- AXI4 response output back to the main crossbar<!-- n5txiq -->
+
+Inside the block it translates the AXI4 requests into an SRAM interface that our primitive RAM wrappers use.<!-- vknoin -->
+It needs to support AXI4 protocol including:<!-- mcykq8 -->
+- Bursts, where the last signal must be indicated correctly.<!-- o02amt -->
+- Response must have the same AXI4 ID as the request<!-- 4t4cew -->
+- Atomic support is *excluded*.<!-- bsi4rc -->
```
martin-velay: This is not covered, how the block behaves when an atomic/exclusive access arrives? What do you reckon, we should update the spec or add a vPlan new item?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
...
+                        Priority: 2
+                        Comment: Corner: partial strobe clears tag
+                    }
+                    subword_read_clears_tag:
+                    {
+                        Reference: raa5pw
+                        Description: Reads of only part of a 64-bit value from a valid cap region return data but with the tag cleared
+                        Node_Type: metric
+                        Priority: 2
+                        Comment: Need to clarify here if the tag is only returned as a clear, or if it permeates as cleared in the tag memory
```
martin-velay: As unclear, we need to request a spec clarification to the designer.
@marnovandermaas, could you clarify the doubt on [this spec line](https://github.com/lowRISC/mocha/pull/667/changes#diff-282c7e37a45f8f9a02c606644a875c3b792222aa0cbd29032d52f3322cb9430bR71)?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
...
+                    Description: Single-beat aligned 64-bit read/write
+                    Node_Type: section_title
+                    write_read:
+                    {
+                        Reference:
+                            '''
+                            vknoin
+                            pgo845
+                            n5txiq
+                            lhjkel
```
martin-velay: Maybe the `lhjkel` ref is not required here as also declared in "execute_from_sram"?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
...
+                        Reference:
+                            '''
+                            kn6exz
+                            af8sx6
+                            u95b14
+                            '''
+                        Description: Both ruser flits of a valid capability read are 1
+                        Node_Type: metric
+                        Priority: 1
+                        Comment: Checks both flits=1 for a valid cap and both flits=0 for an untagged one. af8sx6 asymmetric (one flit 0) is not applicable here. one tag bit per region drives both flits equal
```
martin-velay:
```suggestion
                        Comment: the SRAM invalidate a cap by clearing one or both `ruser` bits, so the two flits are not guaranteed equal by spec. Rather than treat the asymmetric (1,0) case as not-applicable, bind an SVA that on every 2-beat capability read both flits are equal and match the single stored tag bit (ruser_flit0 == ruser_flit1). This proves the one-bit-per-region design never emits a false-valid asymmetric result.
```

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
+{
```
martin-velay, line 6: Missing scenarios?

1. Reset during a burst: `rst_sanity` only checks a clean reset exit. Add a testpoint for `rst_ni` asserted mid-burst (transaction dropped cleanly, no lingering state, clean recovery).
2. Backpressure / ready-throttling: No stress on `aw_ready` / `w_ready` / `ar_ready` deassertion under load. Add a testpoint (and constrained-random throttling) to confirm data is not corrupted on resumption.
3. Richer concurrency: `concurrent_data_tag` covers only one data+cap pairing. Consider multiple outstanding transactions, read/write to the same region in flight, and write-after-write ordering.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
...
+                            vknoin
+                            pgo845
+                            n5txiq
+                            lhjkel
+                            '''
+                        Description: Single-beat 8-byte write followed by readback returns identical data
+                        Node_Type: metric
+                        Priority: 1
+                    }
+                    execute_from_sram:
```
martin-velay: If `axi_sram` ignores `arprot` and returns data identically regardless of the instruction/data bit, there is nothing SRAM-specific to verify here, and real execute-from-SRAM is a core/fetch-path concern. Please either confirm `axi_sram` observes `arprot[2]` in some way, or drop this block-level testpoint (move it to integration/top level). If kept, note in a `Comment` that it is just a read that happens to assert `arprot[2]`.

doc/ref/arch.md
```diff
@@ -1,76 +1,76 @@
...
+- The data width is 64 bits.<!-- jeluga -->
+- The address range and size of the SRAM are defined in the [memory map](#memory-map). Accesses outside this range must return an error, including if only part of the burst is outside the memory range.<!-- u0s8nt -->
```
martin-velay: @marnovandermaas, I think a clarification is required here.
This sentence should maybe be reworded to state that out-of-range error is an SoC/interconnect guarantee (xbar decode -> DECERR), not an `axi_sram` block behavior, or confirm the block is meant to range-check (a design change).

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,379 @@
...
+            {
+                Node_Type: section_title
+                interface_geometry:
+                {
+                    Reference: qohtih
+                    Description: The SRAM block has four ports: clock / reset / AXI4 request in / AXI4 response out
+                    Node_Type: metric
+                    Priority: 1
+                }
+                sram_geometry:
```
martin-velay, line 25: Add a comment to tell from where this comes:
`Comment: SramMemSize 128 KiB, last word 0x1FFF8 and last cap slot 0x1FFF0 are derived from SRAMLength in top_pkg (0x2_0000) and the memmap.svg image, which is untaggable. TODO: point Reference at a real spec tag once the memory map base/size is added as tagged text in arch.md.
`
@marnovandermaas: Would it be possible to have the memory map in Markdown in this file directly instead of the current `svg` file?

martin-velay (commented):
I have some comments, and also I'd like to know if you feel it was a bad experience to create a vPlan based on DVPlan? I think you are among th 1st to do the full exercise (except myself) and I need to get feedback 😃

doc/ref/arch.md
tchilikov-semify: Are atomics planned to be supported? Personally, I think if they are planned in the future, we should add tests to the vplan as V2/V3 milestones, and skip the implementation for now. I was under the impression that there's no support planned.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
tchilikov-semify: agreed and updated

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
tchilikov-semify: updated the comment - It would be good to have the mem map as markdown, so we can tag each line.

doc/ref/arch.md
martin-velay: Yes, it not planned from what I understood too, so you reckon it means that we shouldn't try to see if the design breaks in that case? I'd probably agree with this statement, so I think we probably need to exclude this line. WDYT?

tchilikov-semify: I think thats a valid observation - the sram shouldn't break if an atomic comes through, and I think its worthwhile adding some scenarios to verify this.

A follow up question from my side is: what should the expected behavior be in this case? Should some block outside the SRAM enforce no atomics to SRAM, or should the SRAM accept the request and treat it as a regular read/write, or should the SRAM ignore it all together?

The current behavior is that the SRAM treats atomic requests as normal requests, from my understanding at least.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
tchilikov-semify: moved to integration tests

tchilikov-semify:
Thanks for the feedback, @martin-velay  😃

Regarding DVPlan: I found that the flow and working with the tool was actually quite nice and polished (so far).
I was very fond of the ability to start with a CSV and convert to an HJSON after the fact.

One pain point I have though, is the spec annotation capabilities. A random generated tag means that each time the spec is updated and changed, all of the tags would be re-generated. This would make updating the vplans a bit of a nightmare, since all of the tags would have to be re-mapped.

I think a solution to this could be a per-line / per-sentence hash. This way, the hash only changes if the contents of the line change.

Of course, this sounds simple but it would introduce edge cases (white spaces, punctuation, etc).

I have yet to re-generate or re-map anything, but I anticipate that this would be tedious as it currently stands.

Other than this, it was quite straight-forward!

I plan to map the actual SV testcases with a comment to the tags as well.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
tchilikov-semify: These are all good scenarios to test.

I'll update the vplan accordingly.

doc/ref/arch.md
martin-velay: @marnovandermaas, WDYT?

martin-velay:
> Thanks for the feedback, @martin-velay 😃
> 
> Regarding DVPlan: I found that the flow and working with the tool was actually quite nice and polished (so far). I was very fond of the ability to start with a CSV and convert to an HJSON after the fact.
> 
> One pain point I have though, is the spec annotation capabilities. A random generated tag means that each time the spec is updated and changed, all of the tags would be re-generated. This would make updating the vplans a bit of a nightmare, since all of the tags would have to be re-mapped.
> 
> I think a solution to this could be a per-line / per-sentence hash. This way, the hash only changes if the contents of the line change.
> 
> Of course, this sounds simple but it would introduce edge cases (white spaces, punctuation, etc).
> 
> I have yet to re-generate or re-map anything, but I anticipate that this would be tedious as it currently stands.
> 
> Other than this, it was quite straight-forward!
> 
> I plan to map the actual SV testcases with a comment to the tags as well.

Thanks @tchilikov-semify for the feedback.
You could try to update the spec locally and run again the tagging script, I think only the new untagged lines will get a new tag, the others should remain the same. But if a given line change its content, the tag won't be updated and it's up to the approvers to flag that the vPlan should be updated. This is maybe flaky and we also were thinking about having short hashes instead. I am unsure if we have an opened issues but I think that's something to explore further.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
tchilikov-semify: Updated vplan to include these scenarios as well

doc/ref/arch.md
marnovandermaas: Good question, ideally I think the SRAM would report an error on atomic accesses. If the current SRAM just treats them as normal requests then it is ok to test for that in the DV environment but please open an issue to track that.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
marnovandermaas: Reads don't clear the tag in memory it just returns a zero tag when only part of the data is returned over the bus.

doc/ref/arch.md
marnovandermaas: Agreed that this can be a check done outside the axi_sram. So we can remove that requirement if we are convinced the bus returns this error.

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
marnovandermaas: The SVG is generated from RDL. Can you tag that file? https://github.com/lowRISC/mocha/blob/main/rdl/mocha.rdl
Alternatively we can output a markdown as well from https://github.com/lowRISC/mocha/blob/main/util/rdlgenerator.py

doc/ref/arch.md
tchilikov-semify: Thanks for the input, Marno. I've opened a new issue to track this: #695

util/artefacts.py
```diff
@@ -43,6 +43,13 @@
         "build/rdl/rdl.json",
         "doc/img/memmap.svg",
     ],
+    # rdl: embed the memory map table into the architecture document
+    [
+        "util/rdlgenerator.py",
+        "embed-memory-map-md",
```
marnovandermaas, line 42: I would probably replace the image generation with the markdown generation here so we don't have two very similar processes in the codebase.

util/rdlgenerator.py
```diff
@@ -146,6 +146,101 @@ def gen_memory_map(args) -> None:
...
 
 
+MEMORY_MAP_HEADER = ["Base address", "Top address", "Reserved", "Function"]
+
+"""Comments delimiting the generated memory map table inside a markdown document."""
+MEMORY_MAP_MD_BEGIN = "<!-- BEGIN generated memory map -->"
+MEMORY_MAP_MD_END = "<!-- END generated memory map -->"
+
+
+def memory_map_table(rdljson: dict) -> list[str]:
```
marnovandermaas: This dict doesn't have the sub-types. You can check you've done this correctly using mypy.

doc/ref/arch.md
```diff
@@ -90,7 +90,7 @@ There should be an assertion to notify when writes occur where `wuser` is set to
 There should also be an assertion for `wuser` mismatches, where one part of the capability is marked as valid while another is invalid in the same transaction.<!-- 9a3xf6 -->
 If a portion of the 128-bit aligned region is written it must clear the tag for the whole region including when a partial write strobe is used.<!-- 893tz4 -->
 
-Reads that only read part of a 64-bit value are allowed from valid capability regions, but these should have their tag cleared.<!-- raa5pw -->
+Reads that only read part of a 64-bit value are allowed from valid capability regions, but these reads should have their tag cleared in the response. These reads do not modify the state of the tag in memory.<!-- raa5pw -->
```
marnovandermaas: Each sentence should have its own line.

marnovandermaas (commented):
I've left some reviews on the initial commits (the vplan I'll leave to @martin-velay

doc/ref/arch.md
```diff
@@ -1,76 +1,99 @@
...
 
 ## Memory map
 
-This is the current memory map for Mocha, where the base and top addresses are inclusive, and reserved is the amount of memory reserved for this function:
-
-![Mocha memory map](../img/memmap.svg)
+This is the current memory map for Mocha, where the base and top addresses are inclusive, and reserved is the amount of memory reserved for this function:<!-- mcbp27_x -->
+
+<!-- BEGIN generated memory map -->
+| Base address | Top address | Reserved   | Function      |
```
martin-velay, line 26: The missing tags here don't cause any trouble to DVPlan?

doc/ref/arch.md
```diff
@@ -1,76 +1,99 @@
...
+- Clock input<!-- 947rwh -->
+- Reset input<!-- rrni5j -->
+- AXI4 request input from the main SoC sub-system crossbar<!-- pgo845 -->
+- AXI4 response output back to the main crossbar<!-- n5txiq -->
+
+Inside the block it translates the AXI4 requests into an SRAM interface that our primitive RAM wrappers use.<!-- vknoin -->
+It needs to support AXI4 protocol including:<!-- mcykq8 -->
+- Bursts, where the last signal must be indicated correctly.<!-- o02amt -->
+- Response must have the same AXI4 ID as the request<!-- 4t4cew -->
+- Atomic support is *excluded*. An atomic accesses should return an error.<!-- bsi4rc -->
```
martin-velay, line 79: This is not covered in the vPlan, we should add it?

util/rdlgenerator.py
martin-velay: Need to be removed from here too?

And also need to delete the file itself `doc/img/memmap.svg` from the repo?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+                    concurrent_data_tag:
+                    {
+                        Reference: vknoin
+                        Description: A plain-data write and a capability write issued concurrently both complete and read back correctly
+                        Node_Type: metric
+                        Priority: 2
+                    }
+                    multiple_outstanding_transactions:
+                    {
+                        reference: mcykq8
```
martin-velay:
```suggestion
                        Reference: mcykq8
```

martin-velay: DVPlan will understand, but it's more to have a consistent file. If you prefer you can also change them all in lowercase everywhere (not only for "Reference")

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+                        Priority: 1
+                    }
+                    rst_mid_burst:
+                    {
+                        Reference: rrni5j
+                        Description: DUT is reset during an AXI burst
+                        Node_Type: metric
+                        Priority: 1
+                        Comment: Transaction should be dropped cleanly. The next transaction after an in-flight reset should still be accepted as normal. Would be good to clarification on the expected state of SRAM after a reset (i.e if a reset comes in during a 2-beat capability write)
+                    },
```
martin-velay: Extra comma?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+                    {
+                        Reference: 893tz4
+                        Description: A sub-64-bit (partial write-strobe) write to a tagged slot clears the tag
+                        Node_Type: metric
+                        Priority: 2
+                        Comment: Corner: partial strobe clears tag
+                    }
+                    subword_read_clears_tag:
+                    {
+                        Reference: raa5pw
```
martin-velay: Also add `832lpx` to avoid to have a hole

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+        }
+        Integration:
+        {
+            Description: Verification of axi_sram integrated behind the real Mocha interconnect.
+            Node_Type: section_title
+            Address_map:
+            {
+                Description: Address decode through the real mocha AXI crossbar. Only the SRAM aperture (SRAMBase=0x10000000, 128 KiB) returns OKAY; every other address returns DECERR (the xbar has no default master port; non-SRAM device ports carry axi_err_slv). The scoreboard is decode-aware: in-aperture it checks data/tag, out-of-aperture it expects DECERR and flags any OKAY as a decode bug.
+                Node_Type: section_title
+                sram_reachable:
```
martin-velay, line 408: No "Reference", maybe to `lhfz6r`?

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+                }
+                burst_wrap:
+                {
+                    Reference: u0s8nt
+                    Description: A burst starting in SRAM whose later beats overflow from the address range.
+                    Node_Type: metric
+                    Priority: 3
+                    Comment: AXI4 forbids a burst from crossing a 4 KB boundary, so a compliant master can't issue such a burst. axi_sram itself does not range-check (mem_err_i=0; it masks the address and would silently wrap), so range enforcement rests entirely on the AXI driver being compliant.
+                }
+                sram_boundary:
```
martin-velay, line 431: Same, no "Reference", maybe to `lhfz6r`?

doc/ref/arch.md
```diff
@@ -1,76 +1,99 @@
...
+| 0x40040000   | 0x4004ffff  | 64.0 kiB   | POWER_MANAGER |<!-- vtnnyw -->
+| 0x40050000   | 0x4005ffff  | 64.0 kiB   | ROM_CTRL      |<!-- ov4jpa -->
+| 0x40060000   | 0x4006ffff  | 64.0 kiB   | ENTROPY_SRC   |<!-- s8gkf4 -->
+| 0x41000000   | 0x4100ffff  | 64.0 kiB   | UART          |<!-- heoi9y -->
+| 0x42000000   | 0x4200ffff  | 64.0 kiB   | I2C           |<!-- 94u3qu -->
+| 0x43000000   | 0x4300ffff  | 64.0 kiB   | SPI_DEVICE    |<!-- vniczf -->
+| 0x44000000   | 0x4400ffff  | 64.0 kiB   | TIMER         |<!-- gjk3vu -->
+| 0x45000000   | 0x4500ffff  | 64.0 kiB   | SPI_HOST      |<!-- jkyz51 -->
+| 0x48000000   | 0x4c004003  | 64.0 MiB   | PLIC          |<!-- pfftm1 -->
+| 0x80000000   | 0xbf7fffff  | 1016.0 MiB | DRAM          |<!-- 04smz7 -->
```
martin-velay: I think none of these tags are in the vPlan, if nothing should be verified now add `_x`

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,416 @@
...
+                        Description: DUT comes out of reset cleanly and accepts the first transaction
+                        Node_Type: metric
+                        Priority: 1
+                    }
+                    rst_mid_burst:
+                    {
+                        Reference: rrni5j
+                        Description: DUT is reset during an AXI burst
+                        Node_Type: metric
+                        Priority: 1
```
martin-velay: Just a little note, I am usure it's done on purpose, the priority `0` is the highest one. Bu you can keep it as it is if you like.

doc/ref/arch.md
tchilikov-semify: It seems when I run `dvplan spec_id doc/ref/arch.md` these fields don't get tagged by dvplan. So it seems they don't cause issues.

doc/ref/arch.md
tchilikov-semify: I opened an issue to track the atomics topic: #695. I'll update the vPlan to test that an atomic access should return an AXI error.

doc/ref/arch.md
martin-velay: Oh if it has been done that way it should be fine :)

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,442 @@
...
+                {
+                    Reference: u0s8nt
+                    Description: A burst starting in SRAM whose later beats overflow from the address range.
+                    Node_Type: metric
+                    Priority: 2
+                    Comment: AXI4 forbids a burst from crossing a 4 KB boundary, so a compliant master can't issue such a burst. axi_sram itself does not range-check (mem_err_i=0; it masks the address and would silently wrap), so range enforcement rests entirely on the AXI driver being compliant.
+                }
+                sram_boundary:
+                {
+                    Referece: lhfz6r
```
martin-velay:
```suggestion
                    Reference: lhfz6r
```

hw/top_chip/dv/axi_sram/axi_sram_vplan.hjson
```diff
@@ -0,0 +1,449 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//TODO: Add metrics for each item in the vPlan [#696]
```
martin-velay:
```suggestion
// TODO: Add metrics for each item in the vPlan (issue #696)
```

martin-velay (approved):
Thanks a lot Christian for this PR, I think it's now good to be merged!

PR #670, opened 2026-07-13
[rtl,tagstore] Fix excessive BRAM FPGA utilisation

thomas6785:
The tag controller tag store instantiated prim_ram_1p with parameter .DataBitsPerMask(32'd1)

This allowed write strobes with one bit of granularity, but Xilinx primitives only allow write strobes with byte granularity.

Consequently Vivado was inferring a separate BRAM for each bit of width i.e. 54 BRAMs per instance instead of 1.
There are 8 instances so this PR reduces BRAM18 utilisation by ~200 (from ~300 to ~100)

Also seems to slightly reduce bitstream build time due to reduced contention, on my machine from 39 mins to 30 mins


raylau1:
Please add a patch under `hw/vendor/patches/pulp_axi_llc` with this change so the lint check can pass

raylau1 (approved):
This reduced the tag controller BRAM usage from 204 blocks to 16 blocks! Thanks for fixing this!

marnovandermaas (commented):
Very nice fix Thomas!

PR #681, opened 2026-07-14
[fpga] Add clock constraints for SPI device

### LINKED ISSUE #652: SPI clock pin missing timing constraints
No clock constraint has been added to the SPI sck pin. This means that all paths between the SPI clock and Mocha clock are not timed.

The csb pin is also treated as a clock and should have a clock constraint.

raylau1:
Closes #652.


marnovandermaas (approved):
This is not my area of expertise, but I think this looks reasonable to me. I'll leave it up to you to decide whether to merge it.

PR #687, opened 2026-07-17
[i2c, sw] Minor clean ups

KinzaQamar:


martin-velay (approved):
Thanks Kinza for this cleanup.

I'd only suggest to add a comment to tell the user of `i2c.h` that they must always call `i2c_wait_transfer_finish` after each `i2c_write_bytes`/`i2c_read_bytes`.

Other than that, I just ran the `i2c_host_tx_rx` and the associated `_cheri` test, and everything pass.

marnovandermaas (approved):
Looks good to me, please do verify that the I2C regressions still pass with the expected pass rate.

KinzaQamar:
> please do verify that the I2C regressions still pass with the expected pass rate.

The only test needs checking is `i2c_host_tx_rx` and `i2c_host_tx_rx_cheri` which @martin-velay already checked locally (thanks!). `temperature_sensor` test is checked in CI

> I'd only suggest to add a comment to tell the user of i2c.h that they must always call i2c_wait_transfer_finish after each i2c_write_bytes/i2c_read_bytes.

There is a similar comment already exist here [i2c.h#L121](https://github.com/lowRISC/mocha/blob/349bceda0e54248e78d1d11fcb73c7820051f9dd/sw/device/lib/hal/i2c.h#L121)

PR #691, opened 2026-07-27
[util] FPGA resource utilisation reporting + visualiser utility

thomas6785:
This PR provides utility scripts (TCL, Python, and bash) for generating FPGA resource utilisation.
Resource utilisation is not currently constrained but could be in future - it is also a useful proxy for what is costing build time, and generally useful for building an intuitive view of the design.

TCL generates a text report which is quite difficult to read. I have a Python script to parse this and generate sunburst charts like the below:
image
(You can click on a wedge in this diagram to zoom in and see its children in more detail)

I found this utility handy as a way of intuitively understanding the resource cost of certain blocks

# Some notes
- This introduces Numpy and Plotly to the Python environment. These are both very popular and widely supported, but they are also quite "heavyweight" which may or may not be justified
- The bash script is a little fragile as it needs to be run from the repository root due to relative file paths. I'm not sure of a universally supported way of avoiding this problem
- Currently only generates sunbursts plots but many other visualisations are possible and potentially useful
- Currently only analyses resource utilisation but this flow (open project -> run TCL script -> post process in Python) is plausibly very useful for reporting CDC, DRCs, power, timing results, and maybe even bus latency or other performance criteria
- I would love to see a colour-coded diagram showing the physical implementation on FPGA regions. You can view this in the Vivado GUI but there is no mechanism for exporting as far as I know
- Only reports on the current build - would be nice to get a plot showing change over time to identify any changes that may otherwise go unnoticed
- This would be best implemented as a FuseSoC hook, but the docs on using those are currently three words: "TODO document hooks"

I most likely won't have time to address these issues/ideas as it is my last week with lowRISC, but documenting them here in case anybody finds them useful.


util/utilisation_reporting/report_utilisation.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
marnovandermaas: Using user environment is more portable:
https://github.com/lowRISC/mocha/blob/beb0f769f77b8018a614db1b557efc2105b754b6/util/verible_format.sh#L1

util/utilisation_reporting/sunburst.py
marnovandermaas, line 1: Maybe rename this script "sunburst.py"

util/utilisation_reporting/sunbursts.py
```diff
@@ -0,0 +1,157 @@
...
+    return node.util.total_luts
+
+
+def ff_plus_lut_getter(node: Node):
+    """Return the number of flip-flops and LUT's in this node's own reported utilisation."""
+    return node.util.ffs + node.util.total_luts
+
+
+def bram_getter(node: Node):
+    """Return the number of flip-flops in this node's own reported utilisation."""
```
marnovandermaas: This should be BRAMs right?

util/utilisation_reporting/sunbursts.py
```diff
@@ -0,0 +1,157 @@
...
+
+if __name__ == "__main__":
+    # If we are run as a script, take the first command line argument as a path
+    # to a Vivado `report_utilization -hierarchical` text report, parse it, and
+    # generate two sunburst plots: one for BRAM utilisation and one for FF+LUT
+    # utilisation.
+    import sys
+
+    if len(sys.argv) < 2:
+        print("Usage: python plt.py <path_to_vivado_report_utilisation.txt>")
```
marnovandermaas: The script name should be updated right?

util/utilisation_reporting/parse_report.py
```diff
@@ -0,0 +1,71 @@
...
+    module: str  # module/cell type, e.g. "u_xlnx_mig_7_ddr3"
+    util: Utilisation  # this row's own reported utilisation
+    children: list = field(default_factory=list)
+
+
+def extract_utilisation_table(path: str) -> str:
+    """Thin wrapper that finds the 'utilisation by Hierarchy' table in a raw Vivado
+    report_utilization file and returns just its data rows, stripping the file header,
+    table of contents, and footnotes."""
+    with Path.open(path, encoding="utf8") as f:
```
marnovandermaas:
```suggestion
    with Path(path).open(encoding="utf8") as f:
```

util/utilisation_reporting/parse_report.py
```diff
@@ -0,0 +1,71 @@
...
+    data_end = next(i for i in range(data_start, len(lines)) if lines[i].startswith("+"))
+    return "".join(lines[data_start:data_end])
+
+
+def parse_utilisation_report(path: str) -> Node:
+    """Parse a Vivado `report_utilization -hierarchical` text report into a Node tree."""
+    rows = []
+    for line in extract_utilisation_table(path).splitlines():
+        parts = line.split("|")[1:-1]
+        if len(parts) != 10:
```
marnovandermaas: Would it be useful to have a comment explaining where the magic number 10 comes from? This may change if the Vivado reporting changes right?

util/utilisation_reporting/sunbursts.py
```diff
@@ -0,0 +1,157 @@
...
+            branchvalues=branch_value_style,
+            maxdepth=plot_maxdepth,
+        )
+    )
+
+    # Set some behaviours for interacting with it
+    fig.update_traces(
+        marker={"colors":None},  # force Plotly to regenerate colours
+        leaf={"opacity":1},  # ensures leaves get distinct colours
+        branchvalues="total",  # recommended for hierarchical totals
```
marnovandermaas: Do we need this as well as the setting of it at line 110?

marnovandermaas (commented):
Some minor comments from my end. Any reason why this is still marked as draft? I would be happy to get this into the repository.

marnovandermaas (approved):
I'm happy for this to go in.

PR #693, opened 2026-07-29
Integrate new Ethernet

thomas6785:
This PR replaces the old Ethernet IP with the new version at lowrisc/ethernet.

A smoke test is included which covers:
- accesses most (though not all) Ethernet registers
- sending a series of packets over a loopback and checking the received packets match
- read and write to a small number of PHY registers to validate the MDIO connection
- transmit over TX
- receive over RX (requires you to send a packet from another device)

The last two of these do not include any checking but it is useful at least to ensure they don't time out


thomas6785:
Leaving this as draft for a moment as I would like to try accessing more PHY registers to test PHY features if I get a chance before leaving (though I am relatively confident if we can access *some* PHY registers then the rest should be no different)

sw/cmake/uboot.cmake
```diff
@@ -5,8 +5,8 @@
 # Function that builds U-Boot, creating a target with the given output name.
 function(mocha_uboot UBOOT_NAME)
   # U-Boot repository and tag to use.
-  set(UBOOT_REPOSITORY https://github.com/lowRISC/u-boot)
-  set(UBOOT_TAG mocha-mvp2)
+  set(UBOOT_REPOSITORY /home/thomas6785/proj/now/u-boot)
```
marnovandermaas: We should keep this as a lowRISC repo.

marnovandermaas (commented):
Nice to see the smoke test!

thomas6785:
Marking this as ready for review as I am happy with it - however clang is about the smoketest and I don't have the time or the knowledge of C to address this. As far as I can tell it's only stylistic issues (the test passes fine after all)

Issue #696, opened 2026-08-14
[dv,sram] add metrics for each item in the axi_sram.sv vPlan


tchilikov-semify:
Currently, the items dont have associated metrics (i.e functional, assertion, etc). Add the metrics to the vPlan

PR #697, opened 2026-08-18
[dv,sram] Update SRAM vPlan with Metric Types

tchilikov-semify:
Closes issue #696

### DIFF
--- a/hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
+++ b/hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// TODO: Add metrics for each item in the vPlan (issue #696)
 {
     axi_sram:
     {
@@ -21,6 +20,10 @@
                     Description: The SRAM block should have four ports: clock / reset / AXI4 request in / AXI4 response out
                     Node_Type: metric
                     Priority: 0
+                    Items:
+                    [
+                        assert:TODO_interface_geometry
+                    ]
                 }
                 sram_geometry:
                 {
@@ -34,6 +37,10 @@
                     Node_Type: metric
                     Priority: 0
                     Comment: SramMemSize 128 KiB, last word 0x1FFF8 and last cap slot 0x1FFF0 are derived from SRAMLength in top_pkg (0x2_0000) and 5scstu.
+                    Items:
+                    [
+                        assert:TODO_sram_geometry
+                    ]
                 }
             }
             Data_Path:
@@ -54,6 +61,10 @@
                         Description: DUT should come out of reset cleanly and should accept the first transaction
                         Node_Type: metric
                         Priority: 0
+                        Items:
+                        [
+                            testcase:axi_sram_rst_sanity_test
+                        ]
                     }
                     rst_mid_burst:
                     {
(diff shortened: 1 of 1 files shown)


hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
```diff
@@ -20,6 +19,7 @@
                     Reference: qohtih
                     Description: The SRAM block should have four ports: clock / reset / AXI4 request in / AXI4 response out
                     Node_Type: metric
+                    Metric_Type: assert
```
martin-velay: You need to add everywhere one or multi "Item" otherwise the conversion will fail (I'll fix DVPlan as so far it silently fail). But we expect at least placeholders eventhough you don't know the final name

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
```diff
@@ -32,6 +32,7 @@
                         '''
                     Description: Confirm DUT parameterisation matches spec: AxiDataWidth = 64 (8-byte word), CapSize = 128-bit (16-byte capability), 1 CHERI tag bit per 128-bit aligned region,  SramMemSize = 128 KiB
                     Node_Type: metric
+                    Metric_Type: assert
```
martin-velay: Do you plan to write an assertion in an "initial begin"? Maybe a formal test is more appropriated for these kind of things?

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
```diff
@@ -253,6 +275,7 @@
                             '''
```
martin-velay, line 344: Here you mention a testcase and an assertion, so your Metric_Type should be `assert, testcase`

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
```diff
@@ -304,6 +330,7 @@
...
                         Description: Both ruser flits of a valid capability read should be 1
                         Node_Type: metric
+                        Metric_Type: testcase
```
martin-velay: You mention a testcase but in comment you already mentioned an assertion, both are valid or only one of them?

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
martin-velay: This should probably moved under the section "Tag_write"?

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
tchilikov-semify: I was planning to write this as an assertion inside an initial block, since its just a static check. We could also use formal for this, but since its a simple assertion, I don't know if its worth the effort to set up this test in a separate tool. What do you think?

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
tchilikov-semify: Ah, I didn't know we could make Metric_Type a list - I'll update this!

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
tchilikov-semify: Both, I'll update the Metric_Type

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
tchilikov-semify: Good suggestion, I'll move the cg_tag_read to the "tag_read" section as well.

hw/top_chip/dv/axi_sram/data/axi_sram_vplan.hjson
tchilikov-semify: Added - I re-ran this with `dvplan format -o ...` and it seems the tool accepts it...

PR #716, opened 2026-09-09
[sram] SRAM updates for D1 Signoff

tchilikov-semify:
This PR makes a few updates to the SRAM block to prepare it for D1 signoff.

Primarily:
- It creates a standalone fuseSoC file for the SRAM
- Adds ASSERT_KNOWN assertions on outputs where needed
- fixes an RTL bug within the SRAM that causes the tag array to be 8x smaller than it should be

### DIFF
--- a/hw/top_chip/rtl/axi_sram.sv
+++ b/hw/top_chip/rtl/axi_sram.sv
@@ -158,4 +158,15 @@ module axi_sram #(
     end
   end
 
+  // The read data and tag are deliberately omitted, since access to uninitialized memory reads X
+  `ASSERT_KNOWN(AwReadyKnownO_A, axi_resp_o.aw_ready)
+  `ASSERT_KNOWN(ArReadyKnownO_A, axi_resp_o.ar_ready)
+  `ASSERT_KNOWN(WReadyKnownO_A, axi_resp_o.w_ready)
+  `ASSERT_KNOWN(BValidKnownO_A, axi_resp_o.b_valid)
+  `ASSERT_KNOWN(RValidKnownO_A, axi_resp_o.r_valid)
+  `ASSERT_KNOWN_IF(BKnownO_A, axi_resp_o.b, axi_resp_o.b_valid)
+  `ASSERT_KNOWN_IF(RIdKnownO_A, axi_resp_o.r.id, axi_resp_o.r_valid)
+  `ASSERT_KNOWN_IF(RRespKnownO_A, axi_resp_o.r.resp, axi_resp_o.r_valid)
+  `ASSERT_KNOWN_IF(RLastKnownO_A, axi_resp_o.r.last, axi_resp_o.r_valid)
+
 endmodule
--- a/hw/top_chip/rtl/axi_sram.sv
+++ b/hw/top_chip/rtl/axi_sram.sv
@@ -18,9 +18,10 @@ module axi_sram #(
 );
 
   // Every tag entry can store AxiDataWidth capability tags
-  localparam int unsigned TagBitAddrWidth = AddrWidth - $clog2(top_pkg::CapSizeBits / 8);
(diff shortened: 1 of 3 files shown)


tchilikov-semify:
The tag address width bug was verified on a local branch, using the unmerged AXI driver, and the work-in-progress SRAM verification environment. The directed testcase consisted of an uninitialized sram to which untagged data was written to sram location `0x4000` and a valid capability to sram location `0x0000`. Then, sram location `0x4000` was read back. We expect to see an unset tag, but a set tag was observed on plain data. After the bugfix, this was not the case and the observed data matched the expected.

hw/top_chip/rtl/axi_sram.sv
```diff
@@ -158,4 +158,15 @@ module axi_sram #(
     end
   end
 
+  // The read data and tag are deliberately omitted, since access to uninitialized memory reads X
+  `ASSERT_KNOWN(AwReadyKnownO_A, axi_resp_o.aw_ready)
```
marnovandermaas, line 174: What about aw_valid and aw_addr?

hw/top_chip/rtl/axi_sram.sv
```diff
@@ -158,4 +158,15 @@ module axi_sram #(
...
+  `ASSERT_KNOWN(ArReadyKnownO_A, axi_resp_o.ar_ready)
```
marnovandermaas, line 175: What about ar_valid and ar_addr?

hw/top_chip/rtl/axi_sram.sv
```diff
@@ -18,7 +18,8 @@ module axi_sram #(
 );
 
   // Every tag entry can store AxiDataWidth capability tags
-  localparam int unsigned TagBitAddrWidth = AddrWidth - $clog2(top_pkg::CapSizeBits / 8);
+  localparam int unsigned TagBitAddrWidth = AddrWidth -
+                                            $clog2(top_pkg::CapSizeBits / top_pkg::AxiDataWidth);
```
marnovandermaas: sram_tag_bit_addr below is calculated by shifting right by `$clog2(top_pkg::CapSizeBits / 8)` are you arguing that logic is wrong as well?

hw/top_chip/rtl/axi_sram.sv
```diff
@@ -158,6 +159,11 @@ module axi_sram #(
     end
   end
 
+  // The tag memory must hold one bit for every capability-sized region of the data memory.
+  `ASSERT_INIT(TagRamCoversDataRam_A,
+               2 ** TagBitAddrWidth ==
```
marnovandermaas: Since we use TagAddrWidth in the primitive it might be good to change this to:
`(2 ** TagAddrWidth) * AxiDataWidth`

marnovandermaas (commented):
Some comments from my end. It would be good to add the missing output asserts before merging.

hw/top_chip/rtl/axi_sram.sv
tchilikov-semify: It was my understanding that for the ASSERT_KNOWN_ADDED item on the signoff, these assertions are only required on the outputs of the block. `aw_valid`, `aw_addr`, `ar_valid`, and `aw_addr` are all inputs to the block, so I omitted them. Also, I think it makes sense to omit them, since on a block level if we want to ensure proper behavior of the block, we should guarantee that the outputs of the block are stable, but we cant guarantee that the inputs are stable (of course we could react to it though). If the inputs are unstable, I think that would rather point to a different block misbehaving. What do you think?

marnovandermaas: Of course sorry for the misunderstanding!

marnovandermaas (approved):
Approving only minor change in the assert necessary.

marnovandermaas (approved):
Looks good to me, thank you for the fixes!

PR #720, opened 2026-09-10
[kmac] add missing output known assertions

### LINKED ISSUE #707: [kmac] Add missing ASSERT_KNOWN assertions for output signals
In the kmac block, app_o, entropy_o and idle_o have no ASSERT_KNOWN assertions. We should add them.

tchilikov-semify:
app_o, entropy_o and idle_o had no ASSERT_KNOWN. idle_o is a registered mubi4 and entropy_o is driven from a synchroniser or tied off, so both are always known; app_o holds up unconditionally too.

The vendored source is patched in place and the change is recorded in 0005-Add-missing-output-known-assertions.patch.

kmac_smoke passes 3/3 with Xcelium in both the unmasked and masked configurations.

Closes #707


marnovandermaas (approved):
Looks good to me!

martin-velay (approved):
LGTM

PR #729, opened 2026-09-11
[pwrmgr] add missing output known assertions

tchilikov-semify:
add ASSERT_KNOWN to fetch_en_o, strap_o, low_power_o and esc_rst_rx_o.

The power manager is generated with ipgen, so the change goes in the template and in the generated RTL, and is recorded in 0005_Add_Missing_Output_Assertions.patch.

`dvsim hw/top_chip/ip_autogen/pwrmgr/dv/pwrmgr_sim_cfg.hjson --tool xcelium -i pwrmgr_smoke --reseed 3` passes with 3/3.


marnovandermaas (approved):
Thanks for adding these.

PR #730, opened 2026-09-11
[rstmgr] add missing output known assertion

tchilikov-semify:
Add ASSERT_KNOWN to sw_rst_req_o.

The reset manager is generated with ipgen, so the change goes in the template and in the generated RTL, and is recorded in 0004_Add_Missing_Output_Assertions.patch.

`dvsim hw/top_chip/ip_autogen/rstmgr/dv/rstmgr_sim_cfg.hjson --tool xcelium -i rstmgr_smoke --reseed 3` passes with 3/3.


marnovandermaas (approved):
Thanks this looks good to me.

PR #734, opened 2026-09-14
[axi_xbar] add output known assertions

tchilikov-semify:
Add ASSERT_KNOWN assertions to axi_xbar.


tchilikov-semify:
@marnovandermaas This PR is intended to close the ASSERT_KNOWN field for the D1 signoff, but I'm not 100% convinced that it makes sense to have these assertions bound to the crossbar outputs... What are your thoughts? Is this necessary, or shall we waive it instead?

marnovandermaas:
@tchilikov-semify I think you're right, it might make sense to waive this for the crossbar. It should be sufficient to rely on the assert known outputs of the individual IP blocks. I'll close this pull request and you can waive the item for the sign-off.

